PR #45, opened 2015-01-12
Added a listener to BoxAPIConnection

sarattall:
The BoxAPIConnection object now supports BoxAPIConnectionListeners which
enable the listener to listen to refresh event

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIConnection.java
+++ b/src/main/java/com/box/sdk/BoxAPIConnection.java
@@ -2,6 +2,8 @@ package com.box.sdk;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @sarattall has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -283,7 +287,7 @@ public boolean needsRefresh() {
         boolean needsRefresh;
 
         this.refreshLock.readLock().lock();
-        if (this.expires == 0) {
+        if (this.expires < 0) {
```
gcurtis: Does this change affect anything? I'd rather leave it out if it isn't needed.

gcurtis:
Looks good! I'll merge as soon as the checkstyle errors are fixed.

src/main/java/com/box/sdk/BoxAPIConnection.java
sarattall: We set expires to 0 to force the API connection to restart.

- -1 --> "null", "not set"
- 0  --> "force refresh the next time we need to"
- n  --> "needs refresh in n milliseconds"

sarattall: I see, your `-1` was our `0` and vice versa. Will revert this change

PR #53, opened 2015-01-16
Add UNKNOWN event type

### LINKED ISSUE #50: Event type parsing should be resilient against unknown types
If a new event type is added to the API, then event parsing will fail since there won't be a corresponding `BoxEvent.Type` enum. This should be improved so that the event type can be set to UNKNOWN.

gcurtis:
The UNKNOWN event type is used in cases where the API returns a new type
that the SDK doesn't understand. This prevents an exception from being
thrown while parsing the event JSON.

Fixes #50.


boxcla:

PR #58, opened 2015-01-16
Always re-add event id to cache after removing

### LINKED ISSUE #57: Duplicate events get emitted from EventStream

TLuthra:
Relying on remove to check if an item is already in the ```LinkedHashSet``` doesn't work because we can get events more than just twice.  The first time an ID is added this will work, and the second time the ID is added this will also work, but the third time will fail because we just removed the ID from the hashset in attempt number 2, so the listener gets called every other time.

Fixes #57

### DIFF
--- a/src/main/java/com/box/sdk/LRUCache.java
+++ b/src/main/java/com/box/sdk/LRUCache.java
@@ -14,9 +14,7 @@ class LRUCache<E> {
 
     boolean add(E item) {
         boolean newItem = !this.linkedHashSet.remove(item);
-        if (newItem) {
-            this.linkedHashSet.add(item);
-        }
+        this.linkedHashSet.add(item);
 
         if (this.linkedHashSet.size() >= MAX_SIZE) {
             Iterator<E> it = this.linkedHashSet.iterator();
--- a/src/test/java/com/box/sdk/LRUCacheTest.java
+++ b/src/test/java/com/box/sdk/LRUCacheTest.java
@@ -26,6 +26,16 @@ public class LRUCacheTest {
         assertThat(added, is(false));
     }
 
+    @Category(UnitTest.class)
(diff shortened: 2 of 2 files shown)


boxcla:

gcurtis:
Items always need to be removed and re-added so that they get moved to the back of the queue. Just re-adding an item keeps its position the same. I think that keeping the remove call but removing the `if(newItem)` check and always re-adding the item will fix this. Something like:

```java
boolean newItem = !this.linkedHashSet.remove(item);
this.linkedHashSet.add(item);
```

Also, can you add a unit test for this bug? You can just copy the `addReturnsFalseForExistingItem` test in `LRUCacheTest` to make another test that adds an existing item twice.

TLuthra:
Yea, I fixed that issue, and added in a unit test for this bug.  Everything should be good now.  Also, something to consider, if we use a ```LinkedHashMap``` we would get a little bit of a convenience advantage of it automatically handling the logic.  The ```containsKey(...)``` method would automatically move it to the front of the LRU.  Also, we can then just overwrite the ```removeEldestEntry(Map.Entry)``` method and that might be simpler on the logic side for us (since we did have a bug with the iterator previously).  Not a big thing that must be changed, but maybe something to look at.

Also, I'm unable to run ```gradle check``` locally.  It seems that 11 tests fail with ```java.netBindException: Address in use``` exception.  I think its probably because I may have something else already running on port 8080 which is what is expected.  That kind of slows down the local dev cycle since I have to push every commit to have Travis run the job, then fix it, re-commit, squash, and then push it again.

gcurtis:
That exception is because the unit tests run a local web server for validating requests and mocking responses. We could probably pick a different port to reduce the odds of other people running into that error.

PR #59, opened 2015-01-17
Added a method to use the preflight api

### LINKED ISSUE #47: Preflight Check for Uploads
Implement a pre-flight check to the API endpoint on uploads.  There should also be functionality to opt-in or opt-out so the client can tell the SDK whether or not to try checking.

https://developers.box.com/docs/#files-preflight-check

TLuthra:
The first version of a couple of methods that check the preflight api according to this spec: https://developers.box.com/docs/#files-preflight-check

Closes #47

### DIFF
--- a/src/main/java/com/box/sdk/BoxFolder.java
+++ b/src/main/java/com/box/sdk/BoxFolder.java
@@ -35,7 +35,6 @@ public class BoxFolder extends BoxItem implements Iterable<BoxItem.Info> {
     private static final URLTemplate DELETE_FOLDER_URL = new URLTemplate("folders/%s?recursive=%b");
     private static final URLTemplate FOLDER_INFO_URL_TEMPLATE = new URLTemplate("folders/%s");
     private static final URLTemplate UPLOAD_FILE_URL = new URLTemplate("files/content");
-    private static final URLTemplate CONTENT_URL_TEMPLATE = new URLTemplate("files/%s/content");
     private static final URLTemplate ADD_COLLABORATION_URL = new URLTemplate("collaborations");
     private static final URLTemplate GET_COLLABORATIONS_URL = new URLTemplate("folders/%s/collaborations");
     private static final URLTemplate GET_ITEMS_URL = new URLTemplate("folders/%s/items/");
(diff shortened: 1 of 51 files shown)


boxcla:

src/main/java/com/box/sdk/BoxFolder.java
```diff
@@ -287,6 +288,51 @@ public void rename(String newName) {
     }
 
     /**
+     * Checks if a new version of the file can be successfully uploaded by using the preflight check.
+     * @param  name        the name to give the uploaded file.
+     * @param  fileSize    the size of the file used for account capacity calculations.
+     * @param  id          the id of the file that has a new version being uploaded.
+     */
+    public void preflightUploadCheck(String name, long fileSize, String id) {
```
gcurtis: I'd prefer that this method be added to BoxFile instead so that an ID doesn't need to be passed in.

gcurtis:
Thanks for adding this! I'm wondering if we should call the method `canUpload` instead. It might be a bit better when calling something like `myFolder.canUpload(...)` or `myFile.canUploadVersion(...)`. I'm interested in opinions on this though.

src/main/java/com/box/sdk/BoxFolder.java
TLuthra: Yea, I was debating that myself, but sure I'll move it to BoxFile and just pass in the parent BoxFolder so we can get the parentID from that.

TLuthra:
Yea the name was something else I was debating, I thought something like ```canUpload``` or ```canUploadVersion``` might mislead a developer.  The docs are very blatant that using the preflight endpoint will not be perfectly accurate for all folders, so ```canUpload``` might be true but the upload will fail.  I was worried someone might just check
```
if(myFolder.canUpload()) {
    myFolder.upload()
}
```
instead of still handling the error case on the upload like
```
if(myFolder.canUpload()) {
    try {
        myFolder.upload()
    } catch (BoxAPIException e) {
        // Handle exception correctly.
    }
}
```
But, if thats not that big of an issue or you don't think that to be the case, i'll be glad to change these names to make it sound much better.

PR #65, opened 2015-02-03
Fixes #63 - added metadata support

### LINKED ISSUE #63: Metadata support
This is a proposal for what using metadata might look like in the SDK. It should to cover the basic requirements of creating, updating, and removing metadata on files.

A Metadata class should be added that will support the default "properties" type. In the future, this class can be extended to possibly support other metadata types.

```java
class Metadata extends BoxJSONObject {

    // Create new, empty metadata.
    Metadata() { ... }

    // Can be created from JSON in order to update existing metadata.
    Metadata(String json) { ... }

    // The following methods all return "this" for chaining when building a
    // JSON patch.
    Metadata add(String path, String value)
    Metadata replace(String path, String value)
    Metadata remove(String path)
    Metadata test(String path, String value)

    String get(String path)

    // Returns the patch JSON for updating existing metadata
    String getPatch() { ... }

    // Override BoxJSONObject method to return normal JSON for creating new
    // metadata.
    @Override
    String getPendingChanges() { ... }
}
```

Metadata can then be created, retrieved, updated, or deleted via the BoxFile class.

```java
class BoxFile {
    // ...

    Metadata createMetadata(Metadata) { ... }
    Metadata getMetadata() { ... }
    Metadata updateMetadata(Metadata) { ... }
    void deleteMetadata() { ... }
}
```

An example of what it'd look like to use the SDK for updating the metadata on a file:

```java
BoxFile myFile = new BoxFile(api, "id");
Metadata metadata = myFile.getMetadata()
    .add("/newKey", "new value")
    .test("/fruit", "apple")
    .remove("/fruit")
    .replace("existingKey", "new value");

myFile.updateMetadata(metadata);
```

codyebberson:


boxcla:

src/main/java/com/box/sdk/Metadata.java
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Creates an empty metadata.
+     */
+    public Metadata() {
+        this.values = new JsonObject();
+    }
+
+    /**
+     * Creates a new metadata.
+     * @param values The initial metadata values.
```
gcurtis: Just for consistency, the beginning of param descriptions aren't capitalized.

src/main/java/com/box/sdk/Metadata.java
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public Metadata(JsonObject values) {
```
gcurtis: We try not to expose the minimal-json types in the SDK's API and instead prefer plain JSON strings, such as:

```java
public Metadata(String json) { ... }
```

If other parts of the SDK need this constructor for performance reasons (so that JSON doesn't have to be re-parsed) then it can be made package-private.

src/main/java/com/box/sdk/Metadata.java
```diff
@@ -0,0 +1,163 @@
...
+    public String toString() {
+        return this.values.toString();
+    }
+
+    /**
+     * Converts a JSON patch path to a JSON property name.
+     * @param path The path that designates the key.  Must be prefixed with a "/".
+     * @return The JSON property name.
+     */
+    private String pathToProperty(String path) {
```
gcurtis: Does the properties type support nested JSON objects? This method is okay for now if metadata can only be a string:string key-value.

gcurtis:
Thanks for implementing this! I like the idea of having overloaded methods for default and custom metadata types.

I left a couple comments about some minor changes but overall things look really good. I'm happy with merging after those changes are made.

PR #66, opened 2015-02-03
Fixed custom domain support

codyebberson:
The SDK has partial support for custom domains using getBaseURL/setBaseURL and getBaseUploadURL/setBaseUploadURL.

There were two missing pieces though:

First, the SDK also required getTokenURL/setTokenURL to use a custom domain token endpoint.

Second, the BoxFolder class did not use BoxAPIConnection.getBaseUploadURL, so uploads would not work.

This commit fixes those two issues.

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIConnection.java
+++ b/src/main/java/com/box/sdk/BoxAPIConnection.java
@@ -44,6 +44,7 @@ public class BoxAPIConnection {
     private String userAgent;
     private String accessToken;
     private String refreshToken;
+    private String tokenURL;
     private String baseURL;
     private String baseUploadURL;
     private boolean autoRefresh;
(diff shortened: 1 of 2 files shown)


boxcla:

gcurtis:
Good catch with the file upload bug - we should merge that fix.

However, I'm not sure about making the token URL gettable/settable. I know it may be useful internally for testing purposes, but I don't know if there would ever be a use for it outside of Box. The rationale for making upload/endpoint URLs customizable was so developers could possibly change the API version. For example, if we added a 2.1 API the URL could be changed from api.box.com/2.0/ to api.box.com/2.1/. If a similar case can be made for the tokens endpoint, then I'm okay with making it customizable.

If you split the upload bug fix into a separate PR then we can merge that right away and talk more about the token URL stuff. Also, please format the commit message body be hard-wrapped to around 72 characters. Thanks!

codyebberson:
Internal testing was my personal use case, but there are additional scenarios.  For example, Box maintains https://www.boxenterprise.net/ specifically for customers and regions where box.com is blocked.  Also, some large enterprise customers internal networks only allow https://box.mycompany.com.

codyebberson:
(If that argument is not compelling enough, then I'm happy to split into two PR's.  No worries)

gcurtis:
Ah, okay that makes sense. Sounds good to me, I just wanted to be sure we had a use-case for it.

PR #70, opened 2015-02-06
Fixes #67, Fixes #68, Fixes #69 - Metadata feedback and fixes

### LINKED ISSUE #67: Metadata integration tests fail
When metadata is set in `createMetadataSucceeds()` and `updateMetadataSucceeds()` in `BoxFileTest`, there's no leading slash which causes the key to be set as `oo` instead of `foo`.

```java
uploadedFile.createMetadata(new Metadata().add("foo", "bar"));
```

should be

```java
uploadedFile.createMetadata(new Metadata().add("/foo", "bar"));
```

### LINKED ISSUE #68: Metadata get(String) method should take a path instead of a key
The `Metadata.get(String)` method should take a path so that nested metadata keys can be retrieved. It will also make it consistent with the add, test, replace, and remove methods.

### LINKED ISSUE #69: Metadata class shouldn't throw NPE for missing keys
The `getID()`, `getTypeName()`, and `getParentID()` methods can possibly throw a NPE if an empty metadata object is created. Similarly, the `get(String)` method will throw an exception if a requested key doesn't exist. This is because `this.values.get("some-key")` returns null and then `asString()` gets called on it.

These methods should do null checks before calling `asString()` and return null if the key doesn't exist.

codyebberson:
Now using the "/path" format for all operations, including get().

Handling missing values more gracefully, returning null.

Fixed bug in integration tests.

Added a bunch of unit tests to catch key vs. path problems.

### DIFF
--- a/src/main/java/com/box/sdk/Metadata.java
+++ b/src/main/java/com/box/sdk/Metadata.java
@@ -2,6 +2,7 @@ package com.box.sdk;
 
 import com.eclipsesource.json.JsonArray;
 import com.eclipsesource.json.JsonObject;
+import com.eclipsesource.json.JsonValue;
 
 /**
  * The Metadata class represents one type instance of Box metadata.
@@ -41,7 +42,7 @@ public class Metadata {
      * @return the metadata ID.
      */
     public String getID() {
-        return this.values.get("$id").asString();
+        return this.get("/$id");
     }
 
     /**
@@ -49,7 +50,7 @@ public class Metadata {
      * @return the metadata type.
      */
     public String getTypeName() {
-        return this.values.get("$type").asString();
+        return this.get("/$type");
     }
 
     /**
@@ -57,7 +58,7 @@ public class Metadata {
      * @return the parent object ID.
      */
     public String getParentID() {
-        return this.values.get("$parent").asString();
+        return this.get("/$parent");
     }
 
     /**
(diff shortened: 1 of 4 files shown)


boxcla:

src/main/java/com/box/sdk/Metadata.java
```diff
@@ -142,6 +147,9 @@ public String toString() {
      * @return the JSON property name.
      */
     private String pathToProperty(String path) {
+        if (path == null || !path.startsWith("/")) {
+            throw new IllegalArgumentException("Path must be prefixed with a \"/\".");
```
gcurtis: I like this idea of throwing an IllegalArgumentException for invalid paths.

src/test/java/com/box/sdk/MetadataTest.java
```diff
@@ -0,0 +1,119 @@
...
+    @Category(UnitTest.class)
+    public void testNullPath() {
+        try {
+            new Metadata().add(null, "value");
+            Assert.fail("Expected IllegalArgumentException");
+        } catch (IllegalArgumentException ex) {
+        }
+    }
+
+    @Test
```
gcurtis: You can also do `@Test(expected=IllegalArgumentException.class)` to make it a bit clearer that you're testing that an exception is being thrown. It'll also let you get rid of the try/catch and the `Assert.fail()`.

codyebberson: Yeah...  Using "expected" causes problems for EclEmma (code coverage tool), so I try to avoid it.  If that's your preference, I'm happy to make the change.

gcurtis:
Thanks for fixing these! Just left one suggestion about expecting the `IlegalArgumentException` in the unit test.

PR #72, opened 2015-02-10
Adding support for JDK 1.6

### LINKED ISSUE #62: Bug in BoxUser phone number parsing
The "phone" field is being parsed into the `jobTitle` field in `BoxUser`.

codyebberson:
Per customer request, adding backwards support for Java 1.6.

There should be zero functional changes.

Summary of changes:
* Converted "switch on string" to if/else chains.  There may be a slight performance penalty, but it should be negligible compared to the overall HTTPS request.
* Replaced System.lineSeparator() with System.getProperty("line.separator")
* Replaced SimpleDateFormat "X" with "Z"
* Replaced StandardCharsets with custom helper class
* Replaced Files.readAllBytes with custom helper method

Fixes #62 - correctly parsing user phone field

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'signing'
 
-sourceCompatibility = 1.7
+sourceCompatibility = 1.6
 
 group = 'com.box'
 archivesBaseName = 'box-java-sdk'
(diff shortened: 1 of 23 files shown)


boxcla:

codyebberson:
Oy, Java 6 does not fully support ISO 8601.  I withdraw for now.

http://stackoverflow.com/questions/2201925/converting-iso-8601-compliant-string-to-java-util-date

PR #73, opened 2015-02-10
Adding support for JDK 1.6

### LINKED ISSUE #62: Bug in BoxUser phone number parsing

codyebberson:
Per customer request, adding backwards support for Java 1.6.

There should be zero functional changes.

Summary of changes:
* Converted "switch on string" to if/else chains.  There may be a slight performance penalty, but it should be negligible compared to the overall HTTPS request.
* Replaced System.lineSeparator() with System.getProperty("line.separator")
* Replaced SimpleDateFormat "X" with "Z"
* Added extra support for ISO 8601 formatted date/time strings
* Replaced StandardCharsets with custom helper class
* Replaced Files.readAllBytes with custom helper method

Fixes #62 - correctly parsing user phone field

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIRequest.java
+++ b/src/main/java/com/box/sdk/BoxAPIRequest.java
@@ -8,7 +8,6 @@ import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.ProtocolException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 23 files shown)


boxcla:

src/main/java/com/box/sdk/BoxAPIRequest.java
```diff
@@ -321,7 +321,7 @@ private BoxAPIResponse trySend(ProgressListener listener) {
         HttpURLConnection connection = this.createConnection();
 
         if (this.bodyLength > 0) {
-            connection.setFixedLengthStreamingMode(this.bodyLength);
+            connection.setFixedLengthStreamingMode((int) this.bodyLength);
```
gcurtis: Wow - that's an unfortunate limitation of Java 6. We should be using chunked streaming for all uploads, so I think we should still be okay with large files.

PR #74, opened 2015-02-11
Added event listener for stream position changes

TLuthra:
If we want to be able to start and stop the ALF stream more granularly, we need to know the last stream position that has already been checked.  There is already a way to start an ALF stream at a specific point, but getting the next_stream_position from the request wasn't there.


boxcla:

gcurtis:
Thanks! This looks good.

Issue #82, opened 2015-03-11
Auto renewal of tokens not happening when the system unused for few hours (expired)


tcharles79:
Found that the SDK is not renewing the tokens automatically when the system was unused for few hours. API is responding with 401 error, but new SDK is not handling that to renew it.

However old SDK is renewing the token if it identifies the 401 error for the above scenario.

New SDK does renew the token if it is expiring in near feature. But if it was expired, it is not renewing.

gcurtis:
The SDK should be refreshing if the token has expired. What version/commit hash of the SDK are you using? Also, is this happening when restoring the session by constructing a new BoxAPIConnection using saved tokens (for example, after your program quits and is relaunched)?

tcharles79:
Greg - Not sure which commit, but i downloaded the code on Feb 5 2015. Yes it is happening when i initiate a new connection (new BoxAPIConnection) with the token from database.

gcurtis:
I think this might be a bug related to having an expires timestamp of 0. As a workaround, try calling [`setExpires(long)`](http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxAPIConnection.html#setExpires(long)) after constructing the API connection.

Using the actual expires time from the last session would be ideal, but setting it to any value greater than 0 should fix your issue.

gcurtis:
Hey Charles - there should be a fix for this in place now. Can you let me know if setting the expires timestamp fixed this issue for you? Thanks!

tcharles79:
Hey Greg, I've downloaded the code today and tested this.

When i tested with expired token, the new token pair have been generated. But found that whenever we go for a new BoxAPIConnection, the code is refreshing the tokens even if the token is still valid.

gcurtis:
This is expected if you don't set an expires timestamp. When you create a new API connection with:

```java
BoxAPIConnection api = new BoxAPIConnection(clientID, clientSecret, accessToken,
    refreshToken);
```

the connection has no way of knowing when the access token will expire. To be safe, it will refresh before making the first request in case it has expired. After refreshing, it will remember the new expiration time and only refresh again if your token is about to expire.

If you want to avoid the initial refresh, you can store the expires timestamp along with your tokens, and then set it when you restore the connection:

```java
BoxAPIConnection api = new BoxAPIConnection(clientID, clientSecret, accessToken,
    refreshToken);
api.setExpires(storedExpirationInMilliseconds);
```

mikebell90:
[...]

PR #83, opened 2015-03-11
Refresh access token when expires timestamp is 0

gcurtis:
The `needsRefresh()` method in `BoxAPIConnection` would always return
false if the expires timestamp is 0. A more intuitive behavior is to
always refresh if we don't know when the token is going to expire. This
avoids issues with constructing an API connection and forgetting to call
`setExpires(long)`. For example:

	BoxAPIConnection api = new BoxAPIConnection("id", "secret",
		"storedAccess", "storedRefresh");

	// This call can fail if the stored access token had expired because
	// the expiration timestamp is 0.
	BoxFolder rootFolder = BoxFolder.getRootFolder(api);

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIConnection.java
+++ b/src/main/java/com/box/sdk/BoxAPIConnection.java
@@ -310,13 +310,9 @@ public class BoxAPIConnection {
         boolean needsRefresh;
 
         this.refreshLock.readLock().lock();
-        if (this.expires == 0) {
-            needsRefresh = false;
-        } else {
-            long now = System.currentTimeMillis();
-            long tokenDuration = (now - this.lastRefresh);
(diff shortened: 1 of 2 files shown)


boxcla:

gcurtis:
This is a fix for #82.

gcurtis:
Hey @TLuthra would you mind taking a look at this? I think you guys ran into this issue too (where not setting the expiration timestamp would 404).

I tested this change the best I could, but I'm not sure why there was logic for not refreshing with the expires timestamp is 0. Do you know if there's ever a situation where automatically refreshing when `expires == 0` would break something?

TLuthra:
Yup, we ran into this problem, and we just manually called ```apiConnection.setExpires(-1L)``` to fix it.

I don't think refreshing at ```expires == 0``` would ever break anything, so this looks good!

PR #84, opened 2015-03-12
Exposed URL from a BoxAPIRequest

TLuthra:
To be able to inspect the request a little bit, the first step is to be able to be able to extract out the URL, so I exposed this in a function.


boxcla:

gcurtis:
Sounds good to me. Merge it!

PR #86, opened 2015-03-16
Add getBody() method of BoxAPIRequest

gcurtis:
This allows request interceptors to inspect the body of a request before
it's sent.


boxcla:

PR #87, opened 2015-03-16
Create new enterprise user

gcurtis:
This change adds `createEnterpriseUser()` methods to `BoxUser`. The
basic form of this method creates a new user with a login and name. The
more advanced implementation can take a `CreateUserParams` object which
allows for fully customizing the details about the new user.

For testing, a new `JSONRequestInterceptor` class was added in order to
make it easier to inspect the JSON that would be sent in a request.


boxcla:

PR #88, opened 2015-03-16
Delete enterprise user

gcurtis:
Add a `delete()` method to `BoxUser` that allows administrators to
delete users from their enterprise.


boxcla:

PR #89, opened 2015-03-16
Update a user's info

gcurtis:
This change allows a user's information to be updated by setting fields
in `BoxUser.Info` and then calling `updateInfo()` on a `BoxUser` object.


boxcla:

PR #90, opened 2015-03-16
Get all email aliases for a user

gcurtis:
Add a method called `getEmailAliases()` to `BoxUser` that retrieves all
the email aliases associated with a user.


boxcla:

PR #91, opened 2015-03-16
Add new email alias for a user

gcurtis:
This change adds an `addEmailAlias()` method to `BoxUser` that adds a
new email alias to the user's account.


boxcla:

PR #92, opened 2015-03-17
Delete email alias for a user

gcurtis:
This change adds a `deleteEmailAlias()` method to `BoxUser` that removes
an email alias from their account.


boxcla:

PR #95, opened 2015-03-20
Long download requests

TLuthra:
If a download request gets terminated, or errors the request continues to listen to the input stream since the disconnect isn't called in the exception scenario.  I moved the disconnect in a finally block to make sure that it will be called immediately and not leave the network up and running.


boxcla:

gcurtis:
Looks good to me! Just a minor detail - could you put your pull request description in the actual commit message? It does a good job of explaining what this change does and it'd be helpful to have it in the git log.

PR #97, opened 2015-03-25
Prevent BoxAPIResponse from causing a NullPointerException when using a request interceptor

nealwu:


boxcla:
Hi @nealwu, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

nealwu:
CLA signed

boxcla:
Verified that @nealwu has just signed the CLA. Thanks, and we look forward to your contribution.

src/main/java/com/box/sdk/BoxAPIResponse.java
```diff
@@ -134,19 +134,21 @@ public InputStream getBody(ProgressListener listener) {
      */
     public void disconnect() {
         try {
-            if (this.rawInputStream == null) {
+            if (this.rawInputStream == null && this.connection != null) {
```
gcurtis: I actually think we could just put:

```java
if (this.connection == null) {
    return;
}
```

at the top of the method. I think that makes it a bit clearer that we don't need to do anything if we don't have a connection, and it avoids needing to add additional null checks.

nealwu: Sure, if you think that's good then let's go with that. I just wanted to be a little cautious considering that there was also a change to `this.inputStream` underneath.

gcurtis:
Cool, looks good. Go ahead and merge it!

PR #100, opened 2015-03-31
Closes #94 Get all users in an enterprise.

### LINKED ISSUE #94: Provide options to get all users in an enterprise
In the old SDK, we can get all enterprise users using below code
         
       BoxDefaultRequestObject b1 = new BoxDefaultRequestObject();
	 b1.getRequestExtras().addQueryParam("limit", limit);
	 b1.getRequestExtras().addQueryParam("offset", offset);
        client.getUsersManager().getAllEnterpriseUser(b1, null);

But i don't see the replacement in new SDK for the above use case. 

Please provide. Thanks.

sohtsuka:
Any comments would be highly appreciated (especially on wording).

### DIFF
--- a/src/main/java/com/box/sdk/BoxUser.java
+++ b/src/main/java/com/box/sdk/BoxUser.java
@@ -3,6 +3,7 @@ package com.box.sdk;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 
 import com.eclipsesource.json.JsonArray;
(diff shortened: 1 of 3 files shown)


boxcla:
Verified that @sohtsuka has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/BoxUser.java
```diff
@@ -113,6 +114,41 @@ public static BoxUser getCurrentUser(BoxAPIConnection api) {
...
+    public static Iterable<BoxUser.Info> getAllEnterpriseUsers(final BoxAPIConnection api) {
+        return getAllEnterpriseUsers(api, null);
+    }
+
+    /**
+     * Returns an iterable containing all the enterprise users that matches the filter and specifies which child fields
+     * to retrieve from the API.
+     * @param  api        the API connection to be used when retrieving the users.
+     * @param  filterTerm used to filter the results to only users starting with this string in either the name or the
+     *                    login. Use null if don't want filter.
```
gcurtis: Minor wording suggestion - I think the second sentence should be something like "Can be null to not filter the results."

gcurtis:
Thanks! The code looks really good. I just had a minor suggestion with wording in the docs.

There are also some checkstyle violations that are preventing this from building. If you run `gradle build` you should be able to reproduce them locally.

sohtsuka:
Thanks for reviewing. I have fixed the wording and the checkstyle violations, but when I run `gradle build` I get following errors:

```
com.box.sdk.BoxFolderTest > createFolderSendsRequestWithRequiredFields FAILED
    com.box.sdk.BoxAPIException: Couldn't connect to the Box API due to a network error.

com.box.sdk.BoxGroupTest > createGroupSendsCorrectRequestAndParsesResponseCorrectly FAILED
    com.box.sdk.BoxAPIException: Couldn't connect to the Box API due to a network error.
```

I will investigate later.

gcurtis:
Interesting. I wonder if it's a problem with WireMock. I checked out your branch and built it locally and everything seemed to work. Let me know if you figure out what's causing your build error.

PR #109, opened 2015-04-10
Improve saving and restoring BoxAPIConnections

### LINKED ISSUE #82: Auto renewal of tokens not happening when the system unused for few hours (expired)

### LINKED ISSUE #104: Improve saving and restoring connections
The process of saving and restoring an API connection is pretty complicated and error-prone. The developer needs to know to save the access token, refresh token, expiration time, and last refresh time in order to effectively restore a connection.

One solution is to provide methods that automatically serialize and deserialize the connection's state to and from JSON. The developer would then only need to worry about saving the JSON string. For example:

```java
    ...

    // Restore a saved connection.
    public static BoxAPIConnection restore(String state) { ... }

    // Save the connection's state to a JSON string.
    public String save() { ... }
}
```

`BoxAPIConnectionListener` could also be updated to pass the state string via the `onRefresh()` method.

```java
public interface BoxAPIConnectionListener {

    // The state is the connection's new state after the refresh. It can
    // be saved to restore the connection later.
    void onRefresh(String state);
}
```

gcurtis:
This change adds a `save()` method and a static `restore(String)` method
to BoxAPIConnection. These methods are meant to simplify the process of
saving the state of a connection and then restoring it at a later time.

The `save()` method serializes the state of BoxAPIConnection to a JSON
string which can persisted. The `restore(String)` method takes a JSON
string and reconstructs a BoxAPIConnection using the saved state.

BoxAPIConnectionListener has also had an
`onError(BoxAPIConnection, BoxAPIException)` callback added so that
connection errors can be more easily handled. The `onRefresh()` method
also now receives a BoxAPIConnection object that can be used to save the
newly updated state of the connection.

Closes #104 and fixes #82.

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIConnection.java
+++ b/src/main/java/com/box/sdk/BoxAPIConnection.java
@@ -103,6 +103,42 @@ public class BoxAPIConnection {
     }
 
     /**
+     * Restores a BoxAPIConnection from a saved state.
+     *
+     * @see    #save
+     * @param  state the saved state that was created with {@link #save}.
+     * @return       a restored API connection.
+     */
+    public static BoxAPIConnection restore(String state) {
+        JsonObject json = JsonObject.readFrom(state);
+        String clientID = json.get("clientID").asString();
+        String clientSecret = json.get("clientSecret").asString();
+        String accessToken = json.get("accessToken").asString();
+        String refreshToken = json.get("refreshToken").asString();
+        long lastRefresh = json.get("lastRefresh").asLong();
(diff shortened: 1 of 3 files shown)


boxcla:

PR #111, opened 2015-04-13
Add membership related features

### LINKED ISSUE #77: Retrieve for a User all his groups
Hi,
After a quick overview of the Java Client I was wondering if the API endpoint to retrieve the membership per user is available in the java Client:

https://developers.box.com/docs/#groups-get-all-group-memberships-for-a-user

Because I was not able to find it in BoxUser class.

Cheers

sohtsuka:
Closes #77.

### DIFF
--- a/src/main/java/com/box/sdk/BoxGroupMembership.java
+++ b/src/main/java/com/box/sdk/BoxGroupMembership.java
@@ -1,0 +1,236 @@
+package com.box.sdk;
+
+import java.net.URL;
+import java.text.ParseException;
+import java.util.Date;
+
+import com.eclipsesource.json.JsonObject;
+import com.eclipsesource.json.JsonValue;
+
+/**
+ * Represents a relationship between a user and a group.
+ *
+ * <p>Unless otherwise noted, the methods in this class can throw an unchecked {@link BoxAPIException} (unchecked
+ * meaning that the compiler won't force you to handle it) if an error occurs. If you wish to implement custom error
+ * handling for errors related to the Box REST API, you should capture this exception explicitly.</p>
+ */
+public class BoxGroupMembership extends BoxResource {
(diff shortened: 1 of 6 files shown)


boxcla:
Verified that @sohtsuka has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/BoxGroupMembership.java
```diff
@@ -0,0 +1,235 @@
...
+         * @param  jsonObject the parsed JSON object.
+         */
+        Info(JsonObject jsonObject) {
+            super(jsonObject);
+        }
+
+        /**
+         * Gets the user belonging to the group.
+         *
+         * <p>Note: Only id, name and login fields are available on the initial load.</p>
```
gcurtis: Good thinking documenting this. Can we just rephrase this comment (as well as the others) to say something like:

"Note: the BoxUser.Info returned by this method will only have the ID, name, and login fields populated."

gcurtis:
I've been trying to figure out why `updateMembershipSendsCorrectRequestAndParsesResponseCorrectly()` is failing. When I run it by itself with:

    gradle test --tests *updateMembershipSendsCorrectRequestAndParsesResponseCorrectly

it passes, but when it's run with the other BoxGroupMembership tests, it fails. I'm wondering if this is an issue with WireMock.

If you can't figure what's wrong, then I'm okay with just deleting that test since you also wrote an integration test that covers updating info.

sohtsuka:
Thanks for your comments. 
As I mentioned on the pull request #100, I have been getting different errors since before this commit. All the tests pass if I comment out the following line in BoxAPIRequest.java (line 345):
```
connection.setFixedLengthStreamingMode((int) this.bodyLength);
```
but I can't figure out why the errors pop up.  I will just delete `updateMembershipSendsCorrectRequestAndParsesResponseCorrectly()`.

gcurtis:
Thanks for the contribution!

PR #113, opened 2015-04-14
Parse path collections as BoxFolder.Info

### LINKED ISSUE #105: Path collections should be parsed as a BoxFolder.Info
Right now `BoxItem.getPathCollection()` returns a `List<BoxFolder>`. This means that any additional information that was included in the response (such as the folder's name) gets thrown away. It should parse and return a `List<BoxFolder.Info>` instead.

gcurtis:
When getting info about a folder, its `path_collection` field should be
parsed as a collection of BoxFolder.Info instead of BoxFolder. This way
additional fields (such as the folder's name) can be returned.

Fixes #105.


boxcla:

PR #117, opened 2015-04-15
Remove client ID and secret from connection state

gcurtis:
This change removes the client ID and client secret parameters from the
JSON state blob for BoxAPIConnections. It generally isn't a good
security practice to store your client ID and secret along with
access/refresh tokens, so the restore method was changed to take them as
separate parameters which need to be provided by the application.

There is also a new non-static `restore(String)` method that allows for
a connection state to be restored into an existing connection. This
method helps in situations where the saved state isn't immediately
available when the API connection is constructed.


boxcla:

gcurtis:
@TLuthra - let me know if it looks like this will work for you guys.

Issue #13, opened 2014-11-07
Advanced search


gcurtis:
Add a new search method that takes a `SearchOptions` object that allows for setting all the advanced query parameters accepted by the API.

jdgiotta:
@gcurtis in the integration tests, what is the collaboratorID and how do I obtain it?

jdgiotta:
I've committed a search overload to allow for content_type options, but I can't seem to get integration tests to function. My only form of validation was through additional calls made in the example program.

gcurtis:
@jdgiotta collaboratorID is the ID of another user that will be used when testing collaborations. A user's ID can be found in [their profile settings](https://cloud.app.box.com/settings/profile).

Unfortunately, integration tests are pretty difficult to run because:

1. an enterprise account is required to run all of the tests.
2. you need to manually provide an access and refresh token before each test run.

We're working toward increasing pure unit test coverage by using request interceptors to assert that the correct HTTP requests are being sent. These should be easier to run instead of integration tests.

jdgiotta:
@gcurtis yeah, I just discovered the problem with tokens. Luckily I can obtain new ones fairly easily, Also, I do have an enterprise account and I was able to get most of the tests to run/pass with the exception of download tests, but I think it's a matter of resource configuration on my local project. I'll see if I can't get new integration tests in place before a pull request.

jdgiotta:
Based on my tests, I'm assuming search indexing is delayed.

TLuthra:
Yup, according to Box support, search indexing is a little bit delayed.

https://support.box.com/hc/en-us/articles/202857256-Search-Overview-and-FAQs#search_hdsiw

jdgiotta:
Thanks, @TLuthra Makes for a difficult integration test. I could isolate the test and sleep for 10 minutes LOL

jdgiotta:
Writing an integration test for search is problematic. I can see why one was left out originally.

I isolated my test with a 6 minute sleep, but I took liberty of adding a new task to the Gradle tasks and creating an "IsolatedIntegrationTest" category. I'd hate to fork too greatly in this area.

jdgiotta:
I did not include my tests in this pull requests. I felt the 6 to 10 minute pause would lead to an unwanted experience. If requested, I will include them.

gcurtis:
[...]

PR #120, opened 2015-05-03
BoxFolder.search() Advance Search

jdgiotta:
https://github.com/box/box-java-sdk/issues/13
- Added Overload method for search() to accept String array of content_types. This will allow the user to set where to search within. For example, "tags" or "tags", "comments".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ build/
 .DS_Store
 src/test/config/config.properties
 /target/
+.idea
+*.iml
--- a/src/main/java/com/box/sdk/BoxFolder.java
+++ b/src/main/java/com/box/sdk/BoxFolder.java
@@ -445,6 +445,28 @@ public class BoxFolder extends BoxItem implements Iterable<BoxItem.Info> {
     }
 
     /**
+     * Searches this folder and all descendant folders using a given query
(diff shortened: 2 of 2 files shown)


boxcla:
Verified that @jdgiotta has signed the CLA. Thanks for the pull request!

gcurtis:
Thanks for the contribution!

I think we're going to have to tweak the overloaded method signature for the `search` method before we can merge this. The `content_types` param is one of the many advanced search parameters that the API can take, so I think it would be better to encapsulate all these additional parameters into a separate `AdvancedSearchParams` class. Usage would look something like:

```java
AdvancedSearchParams searchParams = new AdvancedSearchParams()
    .setContentTypes("name", "description")
    .setFileExtensions("png");
folder.search("my query", searchParams);
```

We wouldn't necessarily have to worry about supporting _all_ the different parameters for now. They could always be added later.

.gitignore
```diff
@@ -3,3 +3,5 @@ build/
 .DS_Store
 src/test/config/config.properties
 /target/
+.idea
```
gcurtis: These changes to .gitignore should be left out of this commit. Feel free to submit them as a separate change though.

PR #121, opened 2015-05-06
Handle groups when retrieving and creating collabs

### LINKED ISSUE #116: Groups aren't handled when parsing the accessible_by field in collaborations
Hi!

Playing with the Java SDK  I was trying to retrieve the collaborators for a folder:

```java
Collection<BoxCollaboration.Info> collaborations = parentFolder.getCollaborations();
for(BoxCollaboration.Info collaborationInfo:collaborations) {
    BoxCollaboration.Role role = collaborationInfo.getRole();
    if(isReaderRole(role)) {
        BoxCollaborator.Info accessibleBy = collaborationInfo.getAccessibleBy();
        if(accessibleBy instanceof BoxUser.Info)
            permissionType = USER_PERMISSION_TYPE;
        else if(accessibleBy instanceof BoxGroup.Info)
            permissionType = GROUP_PERMISSION_TYPE;
        aclPermissions.add(permissionType+"-"+accessibleBy.getID());
    }
}
```

Apart the simple code snippets, my problem is that even when the collaborator is a group, I get a BoxUser (a strange instance of this object with really few attributes).

I was expecting a BoxGroup there, am I doing anything wrong?

gcurtis:
Since the "group" type wasn't being handled when creating a new
collaboration, an IllegalArgumentException would be thrown. This fix
handles group types correctly and will successfully create group-based
collaborations.

When retreiving collaborations, the JSON was parsed as a list of users
instead of a list of users or groups. This led to BoxUser.Info objects
being returned with incorrect or unpopulated fields. This fix makes it
so the "type" field is checked for each item in the list, and the JSON
gets parsed into its appropriate type.

Fixes #116.

### DIFF
--- a/src/main/java/com/box/sdk/BoxCollaboration.java
+++ b/src/main/java/com/box/sdk/BoxCollaboration.java
@@ -257,16 +257,12 @@ public class BoxCollaboration extends BoxResource {
                     this.status = Status.valueOf(statusString);
 
                 } else if (memberName.equals("accessible_by")) {
-                    JsonObject userJSON = value.asObject();
+                    JsonObject accessibleByJSON = value.asObject();
                     if (this.accessibleBy == null) {
-                        String userID = userJSON.get("id").asString();
-                        BoxUser user = new BoxUser(getAPI(), userID);
-                        BoxUser.Info userInfo = user.new Info(userJSON);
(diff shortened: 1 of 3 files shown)


boxcla:

PR #122, opened 2015-05-06
Add some additional fields to BoxEvent

gcurtis:
Add getters for the "created_at", "created_by", and "session_id" fields.
There are also getters added for some undocumented fields that may show
up in enterprise events. These fields are "ip_address",
"additional_details", and "accessible_by".

The "additional_details" field doesn't have a known schema, so it has to
be exposed as a raw `JsonObject` (similar to what had to be done for the
"source" field). This may no longer be necessary once the bugs in the
enterprise events stream are fixed.


boxcla:

PR #125, opened 2015-05-16
Added optional functions to use the preflight upload api checks to BoxFolder and BoxFile

### LINKED ISSUE #47: Preflight Check for Uploads

TLuthra:
An updated PR for #59 and closes #47 

(my git history messed up a bunch, so I thought it would be cleaner to just push a new branch)

### DIFF
--- a/src/main/java/com/box/sdk/BoxFile.java
+++ b/src/main/java/com/box/sdk/BoxFile.java
@@ -319,6 +319,32 @@ public class BoxFile extends BoxItem {
     }
 
     /**
+     * Checks if the file can be successfully uploaded by using the preflight check.
+     * @param  name        the name to give the uploaded file or null to use existing name.
+     * @param  fileSize    the size of the file used for account capacity calculations.
+     * @param  parentID    the ID of the parent folder that the new version is being uploaded to.
+     */
+    public void canUploadVersion(String name, long fileSize, String parentID) {
+        URL url = CONTENT_URL_TEMPLATE.build(this.getAPI().getBaseURL(), this.getID());
+        BoxJSONRequest request = new BoxJSONRequest(this.getAPI(), url, "OPTIONS");
+
+        JsonObject parent = new JsonObject();
+        parent.add("id", parentID);
+
+        JsonObject preflightInfo = new JsonObject();
+        preflightInfo.add("parent", parent);
+        if (name != null) {
+            preflightInfo.add("name", name);
+        }
+
+        preflightInfo.add("size", fileSize);
+
+        request.setBody(preflightInfo.toString());
+        BoxAPIResponse response = request.send();
+        response.disconnect();
+    }
+
+    /**
      * Uploads a new version of this file, replacing the current version. Note that only users with premium accounts
      * will be able to view and recover previous versions of the file.
      * @param fileContent a stream containing the new file contents.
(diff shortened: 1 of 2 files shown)


boxcla:

src/main/java/com/box/sdk/BoxFolder.java
```diff
@@ -288,6 +288,30 @@ public void rename(String newName) {
...
+    public void canUpload(String name, long fileSize) {
+        URL url = UPLOAD_FILE_URL.build(this.getAPI().getBaseURL());
+        BoxJSONRequest request = new BoxJSONRequest(this.getAPI(), url, "OPTIONS");
+
+        JsonObject parent = new JsonObject();
+        parent.add("id", this.getID());
+
+        JsonObject preflightInfo = new JsonObject();
+        preflightInfo.add("parent", parent);
+        if (name != null) {
```
gcurtis: The API docs say that a name is required for new files (not for uploading new versions though).

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -319,6 +319,31 @@ public void updateInfo(BoxFile.Info info) {
     }
 
     /**
+     * Checks if the file can be successfully uploaded by using the preflight check.
+     * @param  name        the name to give the uploaded file.
+     * @param  fileSize    the size of the file used for account capacity calculations.
+     * @param  parentID    the id of the parent folder that the new version is being uploaded to.
```
gcurtis: id in the description should be capitalized to ID for consistency.

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -319,6 +319,31 @@ public void updateInfo(BoxFile.Info info) {
     }
 
     /**
+     * Checks if the file can be successfully uploaded by using the preflight check.
+     * @param  name        the name to give the uploaded file.
```
gcurtis: I think it might be worth mentioning that the name can be null in order to use existing name.

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -319,6 +319,31 @@ public void updateInfo(BoxFile.Info info) {
...
+        JsonObject preflightInfo = new JsonObject();
+        preflightInfo.add("parent", parent);
+        if (name != null) {
+            preflightInfo.add("name", name);
+        }
+
+        preflightInfo.add("size", fileSize);
+
+        request.setBody(preflightInfo.toString());
+        request.send();
```
gcurtis: A call to `request.disconnect()` should be made after `send()` to ensure that the connection is cleaned up and reusable for connection pooling.

src/main/java/com/box/sdk/BoxFolder.java
```diff
@@ -288,6 +288,30 @@ public void rename(String newName) {
...
+            preflightInfo.add("name", name);
+        }
+
+        preflightInfo.add("size", fileSize);
+
+        request.setBody(preflightInfo.toString());
+        request.send();
```
gcurtis: Same comment about calling `disconnect()`.

gcurtis:
Thanks! Left some comments for a few minor things.

gcurtis:
I just realized that we should also be returning a boolean for these methods. Otherwise, you'd have to rely on a try-catch to see if the preflight check succeeded.

gcurtis:
Looks good!

PR #128, opened 2015-06-04
Fix OAuth race condition when refreshing

gcurtis:
This change introduces an access token read lock that can be used to
prevent the access token from being refreshed before a request is sent.

When calling `BoxAPIConnection.getAccessToken()`, the access token is
only locked until it is read and returned. This means that a request can
be sent with an invalid access token if a refresh occurred between the
time when `getAccessToken()` was called and the request was sent.

The new `lockAccessToken()` and `unlockAccessToken()` methods allow for
requests to keep the token locked until they're finished being sent.
These methods are package-private for now, since their misuse can easily
cause deadlocks.


boxcla:

ritikm:
Looks good :+1:.

PR #138, opened 2015-06-30
Add ability to set non-global proxy

### LINKED ISSUE #136: Add ability to set non-global proxy
Right now, the only way to set a proxy with the SDK is via the `https.proxyHost` and `https.proxyPort` system properties. This is problematic for applications that can't or don't want to use a proxy for all HTTP requests.

Instead, we can use the [`URL.openConection(Proxy)`](http://docs.oracle.com/javase/7/docs/api/java/net/URL.html#openConnection(java.net.Proxy)) method to allow for a proxy to be set only for HTTP requests sent from the SDK.

Getters and setters should be added to `BoxAPIConnection` which allow for the proxy to be configured.

```java
    // ...

    // Sets a proxy to use when making requests to the API.
    public void setProxy(Proxy proxy) { ... }

    // Sets an authenticated proxy to use when making requests to the API.
    public void setProxy(Proxy proxy, String username, String password) { ... }

    // Gets the proxy that will be used when making requests to the API.
    public Proxy getProxy() { ... }
}
```

bassman531:
Added a Proxy and proxy username/password to BoxAPIConnection.java
Added support for opening the URL connection using this proxy if it exists in BoxAPIRequest.java

Closes #136

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ repositories {
 
 dependencies {
     compile 'com.eclipsesource.minimal-json:minimal-json:0.9.1'
+    compile 'commons-codec:commons-codec:1.10'
     testCompile 'junit:junit:4.11'
     testCompile 'org.hamcrest:hamcrest-library:1.3'
     testCompile 'com.github.tomakehurst:wiremock:1.52'
--- a/src/main/java/com/box/sdk/BoxAPIConnection.java
+++ b/src/main/java/com/box/sdk/BoxAPIConnection.java
@@ -1,6 +1,7 @@
 package com.box.sdk;
 
 import java.net.MalformedURLException;
+import java.net.Proxy;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
@@ -40,6 +41,10 @@ public class BoxAPIConnection {
     // the double-checked lock in getAccessToken(), they must be atomic.
     private volatile long lastRefresh;
     private volatile long expires;
+
+    private Proxy proxy;
+    private String proxyUsername;
+    private String proxyPassword;
 
     private String userAgent;
     private String accessToken;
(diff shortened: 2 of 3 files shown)


boxcla:
Hi @bassman531, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

bassman531:
CLA signed

boxcla:
Verified that @bassman531 has just signed the CLA. Thanks, and we look forward to your contribution.

gcurtis:
This commit is now in #139.

dayasagar:
@gcurtis Is this code commit available in mavenized version ?

dayasagar:
With the latest git code, we tested proxy changes, we observed few issues..

setProxy is available only for BoxAPIConnection, without creating BoxAPIConnection we cant set the proxy

BoxAPIConnection instatiation requires authCode authentication this is again done by sending a request to box, authentication fails again as it cant pass the proxy..

public BoxAPIConnection(String clientID, String clientSecret, String authCode) {
        this(clientID, clientSecret, null, null);
        this.authenticate(authCode);
    }

The below code in BoxAPIRequest trySend method never executes as api is always null..

  if (this.api != null) {
            connection.addRequestProperty("Authorization", "Bearer " + this.api.lockAccessToken());
            connection.setRequestProperty("User-Agent", this.api.getUserAgent());
            if (this.api.getProxy() != null) {
                if (this.api.getProxyUsername() != null && this.api.getProxyPassword() != null) {
                    String usernameAndPassword = this.api.getProxyUsername() + ":" + this.api.getProxyPassword();
                    String encoded = new String(Base64.encode(usernameAndPassword.getBytes()));
                    connection.addRequestProperty("Proxy-Authorization", "Basic " + encoded);
                }
            }
}

I believe this is a deadlock and need to be fixed

gcurtis:
@dayasagar - thanks for reporting this. We're tracking this bug in #142.

This commit isn't available in maven yet. We'll be performing a 1.1.0 release on Monday which will contain this change, as well as the #142 bug fix.

PR #139, opened 2015-06-30
Add ability to set non-global proxy

gcurtis:


boxcla:

PR #144, opened 2015-07-13
Use proxy when authenticating

### LINKED ISSUE #142: BoxAPIConnection should use proxy when authenticating
The `BoxAPIConnection#authenticate(String)` method doesn't use a proxy if one is set. For example:

```java
BoxAPIConnection api = new BoxAPIConnection("clientID", "clientSecret");
api.setProxy(myProxy);
api.authenticate("authCode"); // Authentication doesn't use myProxy.
```

gcurtis:
Update BoxAPIConnection so that the proxy configuration (along with
other connection-level settings) is applied when performing OAuth
requests.

Closes #142.


boxcla:

PR #145, opened 2015-07-13
Update docs around saving/restoring proxy settings

### LINKED ISSUE #143: BoxAPIConnection documentation should mention that proxy configurations aren't saved or restored
Any configured proxy isn't saved or restored with `BoxAPIConnection#save()` or `BoxAPIConnection#restore(String)`. This is intentional because:

1. We don't have a clean way of serializing the `Proxy` object to JSON.
2. If the proxy has a password, it would be unsafe to persist it to the saved state string.

However, the documentation should be clear about this so that developers are aware that the proxy must be manually reconfigured after restoring an API connection.

gcurtis:
Configured proxies aren't saved or restored with
BoxAPIConnection#save() or BoxAPIConnection#restore(String). This is
intentional because:

1. We don't have a clean way of serializing the Proxy object to JSON.
2. If the proxy has a password, it would be unsafe to persist it to the
   saved state string.

This change updates the javadocs so it's clear that proxy settings must
be manually reconfigured after restoring an API connection.

Fixes #143.

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIConnection.java
+++ b/src/main/java/com/box/sdk/BoxAPIConnection.java
@@ -543,6 +543,10 @@ public class BoxAPIConnection {
     /**
      * Saves the state of this connection to a string so that it can be persisted and restored at a later time.
      *
+     * <p>Note that proxy settings aren't automatically saved or restored. This is mainly due to security concerns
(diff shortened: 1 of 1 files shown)


boxcla:

PR #155, opened 2015-10-15
Made public key id mandatory for JWT assertion for App User tokens.

iDeepesh:


boxcla:
Verified that @iDeepesh has signed the CLA. Thanks for the pull request!

Issue #196, opened 2016-01-18
BoxFile.getInfo(BoxFile.All_Fields) returns error if preview is not supported


timsteen:
For several mime types we get an BoxApiException "The API returned an error code: 415" when calling BoxFile.getInfo(BoxFile.All_Fields). 
The detailed message states: "Previews for 'png' files are not yet supported."
The same is true for boxnotes, tiff, 7z, zip, jpeg.

The expectation is that the info for these files can be retrieved even if a preview is not supported.

This was introduced with box-java-sdk 2.0.0.

cburnette:
@gcurtis this is happening because preview_link is a field now but not all files support preview yet. I guess we need to remove it from the all fields array. I can fix this and then I'd like to push 2.1.0

cburnette:
@timsteen as a work around you can remove preview_link from the fields array before you make the call.

timsteen:
Thanks for the quick answer. But there is no "preview_link" in the fields array. There is only a "shared_link" (as in 1.1.0) field and a new expiring_embed_link".

cburnette:
Sorry, it's the expiring_embed_link. I was not near my computer when I wrote that.

timsteen:
Thanks, that helped!

PR #197, opened 2016-01-20
fixes issue #196

cburnette:
removed expiring_embed_link from ALL_FIELDS. Not all file types support preview yet and this is also not a field that should just casually be returned as "info"


boxcla:
Verified that @cburnette has signed the CLA. Thanks for the pull request!

Issue #179, opened 2015-12-30
metadata issue (i think with templates)


cburnette:
-          Custom Meta Data: The key value pairs can only be extracted if the key is known beforehand. There is no possibility to get all existing pairs.

PR #215, opened 2016-03-07
Fix metadata

cburnette:
Fixes issue #179

Added support for specifying the scope parameter when performing metadata operations.  This is required to access any custom metadata types.


boxcla:
Verified that @cburnette has signed the CLA. Thanks for the pull request!

Issue #217, opened 2016-03-15
Any reason not to have BoxGroup.getCollaborations()?


maurercw:
It would be helpful if BoxGroup had a getCollaborations() method.  Any reason not to have one?

Chris

PR #218, opened 2016-03-22
Add method getCollaborations so that collaborations can be returned f…

maurercw:
…rom a group

Addresses issue #217


boxcla:
Verified that @maurercw has signed the CLA. Thanks for the pull request!

tarrencev:
Looks good, thanks for the PR!

Issue #225, opened 2016-04-11
Send non-file parts first in BoxMultipartRequest


klk-:

tarrencev:
Hi @klk-, would you be interested in contributing your fix to this repo?

PR #226, opened 2016-04-12
Issue #225: Send file bytes after JSON

klk-:
This is in accordance with Box API recommendation.

From https://box-content.readme.io/reference#upload-a-file:
"Note that the file bytes should come after the JSON for best
performance."


boxcla:
Verified that @klk- has signed the CLA. Thanks for the pull request!

Issue #198, opened 2016-01-22
EventLog.getEnterpriseEvents fails when it is called with timezone containing "+"


hiroyukitanaka:
com.box.sdk.EventLog.getEnterpriseEvents ends with BoxAPIException when it is called with "+" timezone.

My PC's timezone setting is "(UTC+09:00) Osaka,Sapporo,Tokyo".

I wrote the following code:
```java
BoxAPIConnection api = new BoxAPIConnection("my-developer-token"); 

Calendar cal = Calendar.getInstance();
Date before = cal.getTime();
cal.add(Calendar.DATE, -10);
Date after = cal.getTime();
		
EventLog log = EventLog.getEnterpriseEvents(api, after, before);
```

When I execute the above code, BoxAPIException occurs with the following message.
```
com.box.sdk.BoxAPIException: The API returned an error code: 400
```

I found the following implementation in EventLog.getEnterpriseEvents function.
```java
String afterString = BoxDateFormat.format(after);
String beforeString = BoxDateFormat.format(before);
URL url = ENTERPRISE_EVENT_URL_TEMPLATE.build(api.getBaseURL(), afterString, beforeString);
```

And, I found that `url` has like the following `query` value, when I execute my code on my PC.
```
stream_type=admin_logs&limit=500&created_after=2016-01-12T14:20:18+0900&created_before=2016-01-22T14:20:18+0900
```

I've tried to call Enterprise Events API with the above query, and, I've got 400 response code.
When I tried to URL encode "+" and call the same API, I've got OK response. 
```
stream_type=admin_logs&limit=500&created_after=2016-01-12T14:20:18%2B0900&created_before=2016-01-22T14:20:18%2B0900
```

I think that `afterString` and `beforeString` in EventLog.getEnterpriseEvents function should be URL encoded like:
```java
String afterString = URLEncoder.encode(BoxDateFormat.format(after), "UTF-8");
String beforeString = URLEncoder.encode(BoxDateFormat.format(before), "UTF-8");
```

cburnette:
@gcurtis it looks like the appendParam method on QueryStringBuilder class is encoding the string, but the afterString and beforeString params were added using the build method on URLTemplate class.  That method doesn't appear to do any encoding.

gcurtis:
Yeah, it looks ike the before and after params should be removed from the template and instead added with the `QueryStringBuilder`.

danielandefors:
Maybe duplicate of #131?

PR #228, opened 2016-04-15
Timezone issue

### LINKED ISSUE #131: Unable to pull enterprise events
created_after is beyond one year in the past, API only supports time range starting from one year in the past from now.

Getting the above error irrespective of whatever dates I entered. Even tried with current date as start and end dates, still the same error.

Here's the sample code that I am trying-

        TimeZone tz = TimeZone.getTimeZone("UTC");
        Calendar start = Calendar.getInstance(tz);
        start.add(Calendar.MONTH, -1); // retrieve events for last one month
        Calendar end = Calendar.getInstance(tz);

        try {
            EventLog el = EventLog.getEnterpriseEvents(api, start.getTime(), end.getTime(), BoxEvent.Type.values());
            System.out.println("Total- " + el.getSize());
            for (Iterator<BoxEvent> iterator = el.iterator(); iterator.hasNext();) {
                System.out.println(iterator.next().toString());
            }
        } catch (BoxAPIException ex) {
            ex.printStackTrace();
            System.out.println(ex.getResponse());
        }

And the actual error that I am getting-

{"type":"error","status":400,"code":"bad_request","help_url":"http:\/\/developers.box.com\/docs\/#errors","message":"created_after is beyond one year in the past, API only supports time range starting from one year in the past from now","request_id":"310879445579d0cc0fd33"}

itsmanishagarwal:
Fixed #131 & #198 , by removing the date after and before from ENTERPRISE_EVENT_URL_TEMPLATE and added to queryBuilder.

This way the '+' sign in the timezone will get URLencoded properly.

Fixed the errors.

### DIFF
--- a/src/main/java/com/box/sdk/EventLog.java
+++ b/src/main/java/com/box/sdk/EventLog.java
@@ -7,8 +7,6 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
-import org.apache.commons.lang.time.DateUtils;
-
 import com.eclipsesource.json.JsonArray;
 import com.eclipsesource.json.JsonObject;
 import com.eclipsesource.json.JsonValue;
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @itsmanishagarwal has signed the CLA. Thanks for the pull request!

tarrencev:
Looks good, thanks for the PR!

PR #237, opened 2016-05-06
Advanced search enhance

### LINKED ISSUE #236: Java SDK - Advanced Search Enhancement
Working on the Advanced Search Feature for Java SDK, will be submitting a pull request today . May 6 2016 with my changes.

If this is the wrong type of request feel free to change

rschaller-box:
Add search/advancedSearch

Introduced a new BoxSearch class that uses the /search API endpoint. It also allow Advanced Search Capabilities. 

Closes #236

### DIFF
--- a/src/example/java/com/box/sdk/example/SearchExamplesAsAppUser.java
+++ b/src/example/java/com/box/sdk/example/SearchExamplesAsAppUser.java
@@ -357,17 +357,6 @@ public final class SearchExamplesAsAppUser {
         }
     }
 
-    private static void printSearchResults(Iterable<BoxItem.Info> searchResults) {
-        //Crawl the folder
-        System.out.println("--==Results==--");
-
-        for (BoxItem.Info itemInfo : searchResults) {
-            System.out.println("File Found: " + itemInfo.getName() + ", Owner: " + itemInfo.getOwnedBy().getID());
-        }
-
-        System.out.println("");
-    }
-
     private static void printSearchResults(PartialCollection<BoxItem.Info> searchResults) {
         //Crawl the folder
         System.out.println("--==Results fullResultSize: " + searchResults.fullSize() + "==--");
--- a/src/example/java/com/box/sdk/example/SearchExamplesAsAppUser.java
+++ b/src/example/java/com/box/sdk/example/SearchExamplesAsAppUser.java
@@ -1,0 +1,389 @@
+package com.box.sdk.example;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import com.box.sdk.BoxDeveloperEditionAPIConnection;
+import com.box.sdk.BoxFile;
+import com.box.sdk.BoxFolder;
+import com.box.sdk.BoxItem;
+import com.box.sdk.BoxMetadataFilter;
+import com.box.sdk.BoxSearchParameters;
+import com.box.sdk.BoxUser;
+import com.box.sdk.EncryptionAlgorithm;
+import com.box.sdk.IAccessTokenCache;
+import com.box.sdk.InMemoryLRUAccessTokenCache;
+import com.box.sdk.JWTEncryptionPreferences;
+import com.box.sdk.PartialCollection;
+
+
+
+public final class SearchExamplesAsAppUser {
+
+    private static final String CLIENT_ID = "";
+    private static final String CLIENT_SECRET = "";
+    private static final String USER_ID = "";
+    private static final String PUBLIC_KEY_ID = "";
+    private static final String PRIVATE_KEY_FILE = "";
+    private static final String PRIVATE_KEY_PASSWORD = "";
+    private static final int MAX_DEPTH = 1;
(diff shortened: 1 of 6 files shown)


boxcla:
Hi @rschaller-box, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/main/java/com/box/sdk/BoxFolder.java
```diff
@@ -449,26 +385,25 @@ public void canUpload(String name, long fileSize) {
         URL url = GET_ITEMS_URL.build(this.getAPI().getBaseURL(), BoxFolder.this.getID());
         return new BoxItemIterator(BoxFolder.this.getAPI(), url);
     }
-
     /**
-     * Searches this folder and all descendant folders using a given query.
+     * Function is Deprecated, please use "BoxSearch Class instead".
```
gcurtis: I'd suggest rephrasing this a bit and linking to the BoxSearch class:

    "This method is deprecated, please use the {@link BoxSearch} class instead.

src/main/java/com/box/sdk/BoxSearch.java
```diff
@@ -0,0 +1,55 @@
...
+import com.eclipsesource.json.JsonValue;
+/**
+ * Represents search on Box. This class can be used to search through your box instance.
+ * In addition this lets you take advantage of all the advanced search features.
+ *
+ * <p>Unless otherwise noted, the methods in this class can throw an unchecked {@link BoxAPIException} (unchecked
+ * meaning that the compiler won't force you to handle it) if an error occurs. If you wish to implement custom error
+ * handling for errors related to the Box REST API, you should capture this exception explicitly.</p>
+ */
+public class BoxSearch extends BoxResource {
```
gcurtis: I don't think this class needs to extend BoxResource. BoxResource objects are typically ID'd items that can be returned by the API.

src/main/java/com/box/sdk/BoxMetadataFilterRequest.java
```diff
@@ -0,0 +1,29 @@
+package com.box.sdk;
+import com.eclipsesource.json.JsonArray;
+import com.eclipsesource.json.JsonObject;
+/**
+ * <p>BoxMetadataFilterRequest class was designed to wrap the BoxMetadataFilter close.
+ * Best practice is to use the two in conjuction.</p>
+ *
+ */
+public class BoxMetadataFilterRequest {
```
gcurtis: Does this class need to be exposed to the user? i.e., are there situations where they would need to use this to create the JSON?

If not, we could consolidate this code into the BoxSearchParameters class.

gcurtis:
It looks like a lot blank lines were accidentally removed from `BoxFolder`.

src/main/java/com/box/sdk/BoxSearchParameters.java
```diff
@@ -0,0 +1,427 @@
...
+/**
+ * Used to Setup Box Search Parameters
+ *
+ * <p>Advanced Search support here allows a number of parameters be specified to take full advantage of
+ * box search capabilities. Query parameter is required in all cases except when Metadata templates
+ * searching is being used.</p>
+ *
+ */
+public class BoxSearchParameters {
+    private String query = null;
```
gcurtis: You can omit setting these fields to null and 0. They'll be automatically assigned those as default values.

src/main/java/com/box/sdk/BoxSearchParameters.java
```diff
@@ -0,0 +1,427 @@
...
+     * @param trashContent Can be trashed_only or non_trashed_only.
+     */
+    public void setTrashContent(String trashContent) {
+        this.trashContent = trashContent;
+    }
+    /**
+     * Retrieve the existing BoxMetaDataFilterRequest.
+     * @return this.BoxMetaDataFilterRequest
+     */
+    public BoxMetadataFilterRequest getMdFilters() {
```
gcurtis: What do you think of having this return a List of BoxMetadataFilter objects instead?

src/example/java/com/box/sdk/example/SearchExamplesAsAppUser.java
```diff
@@ -0,0 +1,346 @@
...
+            printSearchResults(searchResults);
+
+        } catch (ParseException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
+        }
+
+    }
+
+    public static void testMetaDataFilter(BoxSearchParameters bsp, BoxSearch bs) {
```
gcurtis: I'd suggest breaking up each section of this method into separate methods. The "test" prefix is also kind of confusing because that's the naming convention for unit tests.

src/main/java/com/box/sdk/BoxSearchParameters.java
```diff
@@ -0,0 +1,427 @@
...
+     */
+    private String printParam(Object obj)    {
+        return obj.toString().replaceAll("[\\[\\]]", "");
+    }
+    /**
+     * Checks String to see if the parameter is null.
+     * @param    paramValue Object that will be checked if null.
+     * @return this.true if the parameter that is being checked is not null
+     */
+    public boolean isNotNull(Object paramValue) {
```
gcurtis: This method is kind of confusing because it does more than just check for null. I'd recommend just using `==` when checking for null objects and changing this method to `isNullOrEmpty(String)` if you need a helper method for strings.

gcurtis: It also looks like this method should be made private.

PR #239, opened 2016-05-09
Advanced search enhance

### LINKED ISSUE #236: Java SDK - Advanced Search Enhancement

rschaller-box:
*Note Implemented Recommended Changes from comments found : *
#237]

Add search/advancedSearch

Introduced a new BoxSearch class that uses the /search API endpoint. It also allow Advanced Search Capabilities.

Closes #236


boxcla:
Verified that @rschaller-box has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/BoxSearchParameters.java
```diff
@@ -318,7 +317,7 @@ private String printParam(Object obj)    {
      * @param    paramValue Object that will be checked if null.
      * @return this.true if the parameter that is being checked is not null
      */
-    public boolean isNotNull(Object paramValue) {
+    private boolean isNullOrEmpty(Object paramValue) {
```
gcurtis: The logic of this method needs to be flipped. Right now it returns true if the object is not null.

src/main/java/com/box/sdk/BoxSearchParameters.java
```diff
@@ -288,15 +291,11 @@ public void setTrashContent(String trashContent) {
...
+     * Retrieve the existing BoxMetaDataFilter.
+     * @return this.BoxMetaDataFilter
      */
-    public BoxMetadataFilterRequest getMdFilters() {
-        BoxMetadataFilterRequest bmfr = new BoxMetadataFilterRequest();
-        for (BoxMetadataFilter bmf : this.mdFilters) {
-            bmfr.addBoxMetadataFilter(bmf);
-        }
-        return bmfr;
+    public List<BoxMetadataFilter> getMdFilters() {
```
gcurtis: I think we should rename this to `getMetadataFilters()` to make the name clearer.

src/main/java/com/box/sdk/BoxMetadataFilter.java
```diff
@@ -0,0 +1,146 @@
...
+    public JsonObject getFiltersList() {
+        return this.filtersList;
+    }
+    /**
+     * Set a filter to the filterList, example: key=documentType, value=special.
+     * @param key the key that the filter should be looking for.
+     * @param value the specific value that corresponds to the key.
+     */
+    public void setFilter(String key, String value) {
+        this.filtersList = new JsonObject();
```
gcurtis: Can this class can only have one filter type set at a time? It looks like `filtersList` is being recreated in each setter.

src/main/java/com/box/sdk/BoxSearch.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Searches all descendant folders using a given query and query parameters.
+     * @param  offset is the starting position.
+     * @param  limit the number of search results CANNONT Exceed 1000.
+     * @param  bsp containing query and advanced search capabilities.
+     * @return a PartialCollection containing the search results.
+     */
+    public PartialCollection<BoxItem.Info> search(long offset, long limit, final BoxSearchParameters bsp) {
```
gcurtis: Rename to `searchRange` to be more consistent with `BoxFolder`. That way we can add another search method later which can do automatic paging.

src/main/java/com/box/sdk/BoxSearchParameters.java
```diff
@@ -0,0 +1,442 @@
...
+     * @param fileExtensions applied as a filter for extensions.
+     */
+    public void setFileExtensions(List<String> fileExtensions) {
+        this.fileExtensions = fileExtensions;
+    }
+    /**
+     * Get the from date filter to specify the when a file was created.
+     * @return this.created date from.
+     */
+    public Date getCreatedAtRangeFromDate() {
```
gcurtis: The names of the range methods are kind of confusing. If you create `DateRange` and `SizeRange` classes, then it'll be a bit clearer.

Usage example:

```java
DateRange range = new DateRange(from, to);
params.setCreatedAtRange(range);
```

rschaller-box:
@gcurtis please review I believe we are good to go :)

amjoesph:
Does this advanced search functionality include these features:
- pagination of records 
- retrieve list of folders/files only on a given folder id

Does the search result would provide the path of a file.

gcurtis:
@amjoesph it doesn't automatically handle paging results yet, but it will still support retrieving items from a specific folder ID.

gcurtis:
Squashed and merged in 11e97cce2ecd56c8dfe81a0a3c1c2fa22317d64a.

PR #247, opened 2016-06-01
Metadata fixes

codyebberson:
* Added getMetadata(template, scope) to specify template scope
* Added Metadata.getKeys() to return the list of keys
* Fixed bug in URL encoder not encoding some characters
* Fixed bug in metadata search incorrectly forcing lower case

### DIFF
--- a/src/main/java/com/box/sdk/Metadata.java
+++ b/src/main/java/com/box/sdk/Metadata.java
@@ -1,5 +1,8 @@
 package com.box.sdk;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import com.eclipsesource.json.JsonArray;
 import com.eclipsesource.json.JsonObject;
 import com.eclipsesource.json.JsonValue;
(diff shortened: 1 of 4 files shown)


boxcla:

src/main/java/com/box/sdk/Metadata.java
```diff
@@ -133,10 +136,29 @@ public String get(String path) {
...
+            return value.toString();
+        }
         return value.asString();
     }
 
     /**
+     * Returns a list of metadata proeprty paths.
+     * @return the list of metdata property paths.
+     */
+    public List<String> getKeys() {
```
tarrencev: Should this be renamed, maybe `getPropertyPaths`? I wouldn't expect it to return JSON patch paths from the cur func name

codyebberson:
@tarrencev Renamed to "getPropertyPaths"

In the future, we may want to consider a "Schema" object that includes types and metadata about metadata.  For example, the client currently treats all fields as strings even though integer fields are allowed.  AFAIK, the API does not provide that functionality today.  We should align with the Metadata team on their roadmap.

tarrencev:
Awesome, lgtm 👍 

Looks like the API supports retrieving a metadata templates schema, https://docs.box.com/reference#get-metadata-schema, we'll have to add support to this sdk for it

PR #253, opened 2016-06-13
Implement file lock/unlock

tarrencev:
Initial attempt at file lock/unlock implementation

### DIFF
--- a/src/main/java/com/box/sdk/BoxFile.java
+++ b/src/main/java/com/box/sdk/BoxFile.java
@@ -622,12 +622,11 @@ public class BoxFile extends BoxItem {
         String queryString = new QueryStringBuilder().appendParam("fields", "lock").toString();
         URL url = FILE_URL_TEMPLATE.buildWithQuery(this.getAPI().getBaseURL(), queryString, this.getID());
         BoxAPIRequest request = new BoxAPIRequest(this.getAPI(), url, "PUT");
-        request.addHeader("Content-Type", "application/json");
 
         JsonObject lockConfig = new JsonObject();
         lockConfig.add("type", "lock");
         lockConfig.add("expires_at", BoxDateFormat.format(expiresAt));
-        lockConfig.add("is_download_prevented", Boolean.toString(isDownloadPrevented));
+        lockConfig.add("is_download_prevented", isDownloadPrevented);
 
         JsonObject requestJSON = new JsonObject();
         requestJSON.add("lock", lockConfig);
--- a/src/main/java/com/box/sdk/BoxFile.java
+++ b/src/main/java/com/box/sdk/BoxFile.java
@@ -6,7 +6,6 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


boxcla:

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -604,6 +607,60 @@ public Metadata createMetadata(String typeName, Metadata metadata) {
...
+        requestJSON.add("lock", lockConfig);
+        request.setBody(requestJSON.toString());
+
+        BoxJSONResponse response = (BoxJSONResponse)request.send();
+
+        JsonObject responseJSON = JsonObject.readFrom(response.getJSON());
+        JsonValue lockValue = responseJSON.get("lock");
+        JsonObject lockJSON = JsonObject.readFrom(lockValue.toString());
+
+        return new BoxLock(lockJSON, this.getAPI());
```
tarrencev: I noticed that we sometimes do a `response.disconnect()`, should I add that to these functions?

gcurtis: `BoxJSONResponse.getJSON()` automatically calls disconnect after reading the entire body of the response. If you're reading the body manually, then you do need to call disconnect afterwards. It ensures that the connection can be reused for keep-alive.

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -604,6 +604,60 @@ public Metadata createMetadata(String typeName, Metadata metadata) {
...
+    public BoxLock lock(Date expiresAt, boolean isDownloadPrevented) {
+        String queryString = new QueryStringBuilder().appendParam("fields", "lock").toString();
+        URL url = FILE_URL_TEMPLATE.buildWithQuery(this.getAPI().getBaseURL(), queryString, this.getID());
+        BoxAPIRequest request = new BoxAPIRequest(this.getAPI(), url, "PUT");
+        request.addHeader("Content-Type", "application/json");
+
+        JsonObject lockConfig = new JsonObject();
+        lockConfig.add("type", "lock");
+        lockConfig.add("expires_at", BoxDateFormat.format(expiresAt));
+        lockConfig.add("is_download_prevented", Boolean.toString(isDownloadPrevented));
```
gcurtis: You can just pass in the boolean here instead of casting it to a String.

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -604,6 +604,60 @@ public Metadata createMetadata(String typeName, Metadata metadata) {
...
+     * Locks a file.
+     * @param expiresAt expiration date of the lock.
+     * @param isDownloadPrevented is downloading of file prevented when locked.
+     * @return the lock returned from the server.
+     */
+    public BoxLock lock(Date expiresAt, boolean isDownloadPrevented) {
+        String queryString = new QueryStringBuilder().appendParam("fields", "lock").toString();
+        URL url = FILE_URL_TEMPLATE.buildWithQuery(this.getAPI().getBaseURL(), queryString, this.getID());
+        BoxAPIRequest request = new BoxAPIRequest(this.getAPI(), url, "PUT");
+        request.addHeader("Content-Type", "application/json");
```
gcurtis: This header will be automatically added if you use `BoxJSONRequest`.

src/main/java/com/box/sdk/BoxFile.java
tarrencev: Ah I see, thanks for the clarification

Issue #150, opened 2015-08-12
File Name Got Truncated


ravikantagarwal:
When I tried to create the file with abc";def then file name got truncated. Also if the file name starts with space than also the leading space got truncated. This issue seems with escaping. I have tried with the box 1.1.0 version.

I have tried to create file with name abc;def and abc"def, these files are successfully created.

I tried to debug the box sdk and found that while writing the filename in the request : Content-disposition header (due to some escaping issue with the file name) the file name got truncated. Please refer(writePartHeader method of BoxMultipartRequest).

I also tried to encode the filename using UTF-8 but that also does not work

Can you please let me how to escape these special character in case of box.

gcurtis:
Thanks for reporting this. You're correct that the filename is being incorrectly truncated for names with `"` and `;`. The `uploadFile()` method needs to be updated to specify the filename in JSON instead of the Content-Disposition header.

Filenames with leading or trailing spaces aren't supported by Box, so those are correctly truncated.

ravikantagarwal:
No in box UI we can create a file with leading space, though on box UI the leading space is not visible but when we try to rename the folder it displays the leading space, even if we fetch the name of file using sdk that filename also contains leading space

gcurtis:
Sorry, I misread the docs. Leading spaces should be permitted but trailing spaces are not.

PR #256, opened 2016-06-16
Fix filename truncation #150

tarrencev:
Address #150 by specifying the filename (and parent id) in JSON instead of the Content-Disposition header

@gcurtis @codyebberson @dmaynard

### DIFF
--- a/src/main/java/com/box/sdk/BoxMultipartRequest.java
+++ b/src/main/java/com/box/sdk/BoxMultipartRequest.java
@@ -5,6 +5,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLEncoder;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


boxcla:

src/main/java/com/box/sdk/BoxMultipartRequest.java
```diff
@@ -120,7 +120,7 @@ protected void writeBody(HttpURLConnection connection, ProgressListener listener
                 this.writeOutput(entry.getValue());
             }
 
-            this.writePartHeader(new String[][] {{"name", "filename"}, {"filename", this.filename}},
+            this.writePartHeader(new String[][] {{"name", "file"}, {"filename", this.filename}},
```
codyebberson: Is there a reason for "name=file" in the Content-Disposition header?  AFAIK, it's not part of the standard.  Maybe legacy support?  If no good reason, then perhaps we should remove it.

Also, it appears the writePartHeader still has escaping problems:
https://github.com/box/box-java-sdk/blob/master/src/main/java/com/box/sdk/BoxMultipartRequest.java#L177-L182

Line 181 in particular is directly writing user-submitted data to a HTTP header.  That could be abused.  We should URL-encode the value.

See:  http://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http/6745788#6745788

codyebberson:
Minor nits:
* The encode method could be static, so you don't need to create a new QueryStringBuilder
* We could also remove the encode method entirely and use Java's built in [URLEncoder class](http://docs.oracle.com/javase/6/docs/api/java/net/URLEncoder.html)

Otherwise lgtm, +1

tarrencev:
I switched it to using `URLEncode` earlier :) Makes sense to move `QueryStringBuilder` to that too but I'll leave that for another PR

PR #264, opened 2016-06-21
Search tests and fixes

tarrencev:
Adding unit tests for search functionality (Decided against any integration tests due to the search indexing delay). 
Removed the `keyword` param which, as far as I can tell, is not valid.
Added a check to ensure either a `query` or a `metadata filter` is provided when searching.
Converted `BoxDateFormatter` to use UCT time

@codyebberson @gcurtis @rschaller-box

### DIFF
--- a/src/main/java/com/box/sdk/BoxSearchParameters.java
+++ b/src/main/java/com/box/sdk/BoxSearchParameters.java
@@ -254,6 +254,8 @@ public class BoxSearchParameters {
             if (((String) paramValue).trim().equalsIgnoreCase("")) {
                 isNullOrEmpty = true;
             }
+        } else if (paramValue instanceof List) {
+            return ((List) paramValue).isEmpty();
         }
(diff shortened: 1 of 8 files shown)


boxcla:

src/main/java/com/box/sdk/BoxSearchParameters.java
```diff
@@ -296,24 +273,43 @@ private JsonArray formatBoxMetadataFilterRequest() {
...
+     * Concat a List into a CSV String.
+     * @param list list to concat
+     * @return csv string
+     */
+    private String listToCSV(List<String> list) {
+        String csvStr = "";
+        for (String item : list) {
+            csvStr += "," + item;
+        }
+        return csvStr.substring(1);
```
codyebberson: This will throw exception on empty list.

codyebberson:
LGTM +1

PR #267, opened 2016-06-24
Transactional Auth Flow

tarrencev:


boxcla:

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /**
+     * Request a scoped transactional token for a particular resource.
+     * @param accessToken application access token.
+     * @param scope scope of transactional token.
+     * @param resource resource transactional token has access to.
+     * @return a BoxAPIConnection which can be used to perform transactional requests.
+     */
+    public static BoxAPIConnection getTransactionConnection(String accessToken, String scope, String resource) {
```
tarrencev: We might want to add a helper method for creating resource URL's, the possibilities today are
`https://api.box.com/2.0/files/{file_id}` and `https://api.box.com/2.0/folders/{folder_id}`, so the only info we need to construct it is the type (`file` or `folder`) and the resource id

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,161 @@
...
+     * @param  clientID     the client ID to use when refreshing the access token.
+     * @param  clientSecret the client secret to use when refreshing the access token.
+     * @param  accessToken  an initial access token to use for authenticating with the API.
+     * @param  refreshToken an initial refresh token to use when refreshing the access token.
+     * @throws BoxAPIException Box Transactional API does not support creating an connection with these params
+     */
+    public BoxTransactionalAPIConnection(String clientID, String clientSecret, String accessToken,
+                                            String refreshToken) {
+        super(null);
+        throw new BoxAPIException("This constructor is not available for BoxTransactionalAPIConnection.");
```
gcurtis: I know that `BoxDeveloperEditionAPIConnection` throws BoxAPIException for unsupported constructors, but I think it's more common to throw `UnsupportedOperationException` for unsupported methods.

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Disabling an invalid constructor for Box Developer Edition.
+     * @param  clientID     the client ID to use when exchanging the auth code for an access token.
+     * @param  clientSecret the client secret to use when exchanging the auth code for an access token.
+     * @param  authCode     an auth code obtained from the first half of the OAuth process.
+     * @throws UnsupportedOperationException Box Transactional API does not support creating
+     *         a connection with these params
+     */
+    public BoxTransactionalAPIConnection(String clientID, String clientSecret, String authCode) {
+        super(null);
```
priyajeet: What's the super(null) doing? Is there even a super method in the parent class? Why not just throw directly.

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public void authenticate(String authCode) {
+        throw new UnsupportedOperationException(
+                "BoxTransactionalAPIConnection does not support the authenticate method."
+        );
+    }
+
+    /**
+     * BoxTransactionalAPIConnection can never refresh.
+     * @return true always.
```
priyajeet: false always

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Constructs a new BoxTransactionalAPIConnection that authenticates with an access token.
+     * @param  accessToken a transactional auth access token.
+     */
+    public BoxTransactionalAPIConnection(String accessToken) {
+        super(accessToken);
+        super.setAutoRefresh(false);
+    }
+
+    /**
```
gcurtis: I'd change these docs to just say that the constructors are not used.

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
priyajeet: I think you add @Override annotation to overriden classes. http://docs.oracle.com/javase/6/docs/api/java/lang/Override.html
https://docs.oracle.com/javase/6/docs/technotes/tools/solaris/javadoc.html#@inheritDoc

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,166 @@
+package com.box.sdk;
+
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLEncoder;
+
+import com.eclipsesource.json.JsonObject;
```
priyajeet: Unsure if java 6 has JsonObject natively. I used iirc JSONObject from org.json

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,166 @@
...
+                    "An error occurred while attempting to encode url parameters for a transactional token request"
+            );
+        }
+
+        BoxAPIRequest request = new BoxAPIRequest(apiConnection, url, "POST");
+        request.shouldAuthenticate(false);
+        request.setBody(urlParameters);
+
+        BoxJSONResponse response = (BoxJSONResponse) request.send();
+        JsonObject responseJSON = JsonObject.readFrom(response.getJSON());
```
priyajeet: ParseException - if the input is not valid JSON
java.lang.UnsupportedOperationException - if the input does not contain a JSON object

Need to catch

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
priyajeet: BTW, make the constructors private. Thats how you prevent someone new-ing up

priyajeet: In fact, why add them at all?

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
gcurtis: I like this idea. I wonder if we could eventually do something like:

```java
BoxFile myFile = new BoxFile(api, "1234");
BoxAPIConnection transactionConn = myFile.getTransactionConnection(String scope);
```

priyajeet: preview is being added.
https://api.box.com/2.0/preview/{file_id}

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
tarrencev: Yeah gonna stick with the current lib, we can migrate at a later point if it makes sense. I don't think the JSON implementation is exposed externally, so shouldn't be too tough

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
tarrencev: I think it is necessary to keep them public to override the base classes constructors

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
tarrencev: done

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
tarrencev: @gcurtis I like that too, although I'm somewhat hesitant to litter the transactional auth code throughout the SDK, especially since its primary use case is api calls from the browser. Imo, there isn't a whole lot of utility for scoped tokens on the server side, you may as well just use the fully privileged token for the call, and avoid making the downscope api call

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
tarrencev: I've just been following the patterns used throughout the SDK, and don't think there is a try/catch around this anywhere else

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
priyajeet: In Java (most languages?) constructors are never inherited or overriden. Which is why you have to call super() manually to chain constructors. So if you don't want BoxTransactionalAPIConnection being new-ed up with certain arguments, just don't add a constructor.

tarrencev: Ah I see, we should remove them from the app user connection too then

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
priyajeet: See comment below

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
```diff
@@ -0,0 +1,130 @@
...
+            url = new URL(apiConnection.getTokenURL());
+        } catch (MalformedURLException e) {
+            assert false : "An invalid token URL indicates a bug in the SDK.";
+            throw new RuntimeException("An invalid token URL indicates a bug in the SDK.", e);
+        }
+
+        String urlParameters;
+        try {
+            urlParameters = String.format("grant_type=%s&subject_token=%s&subject_token_type=%s&scope=%s", GRANT_TYPE,
+                    accessToken, SUBJECT_TOKEN_TYPE, URLEncoder.encode(scope, "UTF-8"));
```
codyebberson: accessToken not encoded.  Maybe doesn't matter, because everything will fail if it's invalid, but encoding seems like good hygiene.

You could replace the String.format and try/catch by simply using QueryStringBuilder.  Might be cleaner.

codyebberson:
As discussed with @tarrencev, it feels slightly unorthodox for this to be a subclass of BoxAPIConnection.    Authentication should be relatively independent of most API calls.  Ideally, all different types of authentication generate an access token, and everything is the same after that.

On the other hand, as @tarrencev pointed out, we've already established this pattern with BoxDeveloperEditionAPIConnection, so perhaps this is a matter of consistency.

Other than that design decision, this looks good to me.

src/main/java/com/box/sdk/BoxTransactionalAPIConnection.java
tarrencev: I'll add the encode to accessToken, my thinking was that everything will fail but agree it is good hygiene. The reason I'm not using the QueryStringBuilder is because the API doesn't want GRANT_TYPE and SUBJECT_TOKEN_TYPE to be url encoded =/

PR #269, opened 2016-06-28
Removing Redundant Throws from CheckStyle

dmaynard:
RedundantThrows was removed with version 6.2 (see https://github.com/checkstyle/checkstyle/issues/473)

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -169,7 +169,7 @@
         <module name="IllegalInstantiation"/>
         <module name="InnerAssignment"/>
         <module name="MissingSwitchDefault"/>
-        <module name="RedundantThrows"/>
+        <!-- module name="RedundantThrows"   RedundantThrows was removed with version 6.2  -->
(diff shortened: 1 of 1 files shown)


boxcla:

config/checkstyle/checkstyle.xml
```diff
@@ -169,7 +169,7 @@
         <module name="IllegalInstantiation"/>
         <module name="InnerAssignment"/>
         <module name="MissingSwitchDefault"/>
-        <module name="RedundantThrows"/>
+        <!-- module name="RedundantThrows"   RedundantThrows was removed with version 6.2  -->
```
tarrencev: Can we just remove this?

tarrencev:
minor comment, otherwise lgtm 👍

config/checkstyle/checkstyle.xml
dmaynard: Yes  we could.  I left in in and commented because this used to be supported, and someone might wonder why it is no longer included.

PR #272, opened 2016-07-05
Added metadata scope overrides

codyebberson:
Added BoxFile metadata overrides for non-standard metadata scope.

Metadata templates can be "global" or "enterprise".  The SDK's initial metadata implementation assumed that the default template was "global", and that all other templates are "enterprise".  That assumption is not valid (although does cover most common cases).

I added the equivalent override for getMetadata() in a previous commit.  Now adding support for createMetadata() and deleteMetadata().  The updataMetadata() method does not need an override, because it pulls scope from the Metadata parameter.


boxcla:

PR #282, opened 2016-07-08
Feature / Web Hooks

stanislav-dvorscak:
Implementation for Web Hooks


boxcla:
Hi @stanislav-dvorscak, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

build.gradle
```diff
@@ -3,6 +3,7 @@ apply plugin: 'jacoco'
 apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'signing'
+apply plugin: 'eclipse'
```
gcurtis: What does this plugin do?

build.gradle
```diff
@@ -5,7 +5,7 @@ apply plugin: 'maven'
 apply plugin: 'signing'
 apply plugin: 'eclipse'
 
-sourceCompatibility = 1.6
+sourceCompatibility = 1.7
```
gcurtis: Switching to Java 7 isn't a change that we can make lightly. This would need to happen as part of a major version bump.

src/main/java/com/box/sdk/internal/package-info.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * <b>Not a part of Public API - the backward-compatibility is not maintained.</b><br />
+ * Whole 'internal' package is not a part of the public API and package (and all sub-packages and their classes) are determined only for SDK
+ * implementation.<br />
+ */
+package com.box.sdk.internal;
```
gcurtis: Can we just make these classes package-private to avoid adding another package?

src/main/java/com/box/sdk/internal/utils/CollectionUtils.java
```diff
@@ -0,0 +1,65 @@
...
+     *            type of result
+     * @param <T_Source>
+     *            type of source
+     * @param source
+     *            for mapping
+     * @param mapper
+     *            element mapper
+     * @return mapped source
+     */
+    public static <T_Result, T_Source> List<T_Result> map(Collection<T_Source> source, Mapper<T_Result, T_Source> mapper) {
```
gcurtis: I'm not sure about this method. It really isn't doing much, and an inline foreach loop is much more readable in my opinion.

src/main/java/com/box/sdk/BoxAPIRequest.java
```diff
@@ -51,6 +51,18 @@
 
     /**
      * Constructs an unauthenticated BoxAPIRequest.
+     *
+     * @param url
+     *            the {@link URL} of the request
+     * @param method
+     *            the {@link HttpMethod} of the request
+     */
+    public BoxAPIRequest(URL url, HttpMethod method) {
```
gcurtis: This is kind of redundant with the other constructor that takes a string.

config/checkstyle/checkstyle.xml
```diff
@@ -124,7 +124,7 @@
         <!-- Checks for Size Violations.                    -->
         <!-- See http://checkstyle.sf.net/config_sizes.html -->
         <module name="LineLength">
-            <property name="max" value="120"/>
```
gcurtis: Please avoid changing the style guidelines.

src/main/java/com/box/sdk/BoxResourceIterable.java
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Common implementation for paging support.
+ *
+ * @author Stanislav Dvorscak
+ *
+ * @param <T>
+ *            type of iterated entity
+ */
+public abstract class BoxResourceIterable<T> implements Iterable<T> {
```
gcurtis: You can use `JSONIterator` which already has a lot of this functionality. See [`BoxGroupIterator`](https://github.com/box/box-java-sdk/blob/master/src/main/java/com/box/sdk/BoxGroupIterator.java) as an example of a paging iterator.

gcurtis:
We'd prefer to avoid merging master back into the feature branch, and instead just rebase.

src/main/java/com/box/sdk/BoxWebHook.java
```diff
@@ -0,0 +1,685 @@
...
+     *            {@link BoxItem} web resource
+     * @param address
+     *            {@link URL} where the notification should send to
+     * @param triggers
+     *            events this {@link BoxWebHook} is interested in
+     * @return created {@link BoxWebHook}
+     *
+     * @see #create(BoxItem, URL, Set)
+     */
+    public static BoxWebHook.Info create(BoxItem target, URL address, BoxWebHook.Trigger... triggers) {
```
sekhridhiraj: In future there might be support for other resources as target. E.g. User/Enterprise.
We should make it generic, that makes adding other resource types as target easier.

src/main/java/com/box/sdk/BoxWebHook.java
```diff
@@ -0,0 +1,685 @@
...
+         * Triggered when a {@link BoxFolder} gets deleted.
+         */
+        FOLDER_DELETED("FOLDER.DELETED", TargetType.FOLDER),
+
+        // BoxFile related triggers.
+
+        /**
+         * Triggered when a {@link BoxFile} gets uploaded.
+         */
+        FILE_UPLOADED("FILE.UPLOADED", TargetType.FILE),
```
sekhridhiraj: FILE.UPLOADED is supported on target type folder (not file). Trigger when a file is uploaded in target folder.

src/main/java/com/box/sdk/BoxWebHook.java
```diff
@@ -0,0 +1,685 @@
...
+
+        /**
+         * Triggered when a {@link BoxFile} gets copied.
+         */
+        FILE_COPIED("FILE.COPIED", TargetType.FILE),
```
sekhridhiraj: Other FILE.* and COMMENT.* triggers should be available on both target type file and folder.

src/main/java/com/box/sdk/BoxWebHookSignatureVerifier.java
```diff
@@ -0,0 +1,255 @@
...
+     * @param primarySignatureKey
+     *            primary signature key for web-hooks (can not be null)
+     * @param secondarySignatureKey
+     *            secondary signature key for web-hooks (can be null)
+     * @throws IllegalArgumentException
+     *             primary key can not be null
+     */
+    public BoxWebHookSignatureVerifier(String primarySignatureKey, String secondarySignatureKey) {
+        if (primarySignatureKey == null) {
+            throw new IllegalArgumentException("At least primary signature key must be provided!");
```
sekhridhiraj: I think we should check that either primarySignatureKey or secondarySignatureKey must be present.
There might be scenario where primarySignatureKey is not configured but secondary is configured.

build.gradle
stanislav-dvorscak: It adds eclipse tasks / support, for building eclipse structure.

src/main/java/com/box/sdk/internal/package-info.java
stanislav-dvorscak: Yes, it is, but it is conventional to have multiple packages within Java. Not to have 71 classes within single package.
Do you have any concrete requirement for single package solution?

config/checkstyle/checkstyle.xml
stanislav-dvorscak: I have reverted checkstyle alteration, anyway, maybe it should be considered, if 120 characters is sufficient for nowadays displays.

src/main/java/com/box/sdk/BoxResourceIterable.java
stanislav-dvorscak: According to the Web Hook documentation, Web Hooks paging is based on "next", but JSONIterator is based on "limit" and "offset". Because of this, I have created new implementation and I think paging should be only one, maybe JSONIterator can be deprecated, because limit and offset is not efficient way for paging. Other reason is generic implementation, which avoid duplicate code: isNext. And in case of moving forward to Java 8, it will be more easy for implementation.

src/main/java/com/box/sdk/BoxWebHook.java
stanislav-dvorscak: OK, I have changed implementation to be more generic.

config/checkstyle/checkstyle.xml
wladimiiir: I agree, 120 is way too low for current displays and the need for wrapping single line to multiple lines in most cases leads to worse code readability.

src/main/java/com/box/sdk/BoxAPIRequest.java
wladimiiir: Clean code suggests we should avoid String types and use Enum in such cases to follow fail early principle (during the compilation). I would lean to deprecation of the constructor with String. On the other hand, these kind of "issues" could be handled during the 'Box Java SDK feedback and improvement process' :)

src/main/java/com/box/sdk/BoxWebHook.java
```diff
@@ -0,0 +1,686 @@
...
+     * @throws IllegalArgumentException
+     *             in case of unsupported {@link BoxResource}: only {@link BoxFolder} or {@link BoxFile} is supported.
+     */
+    private static TargetType type(BoxResource target) {
+        if (target instanceof BoxFolder) {
+            return TargetType.FOLDER;
+        } else if (target instanceof BoxFile) {
+            return TargetType.FILE;
+        } else {
+            throw new IllegalArgumentException(String.format("Unsupported BoxItem, expected: '%s' but it was '%s': ",
```
sekhridhiraj: You might want to use "Unsupported target resource..." instead of "Unsupported BoxItem" as the exception message.

src/main/java/com/box/sdk/BoxWebHook.java
stanislav-dvorscak: thanks, fixed

src/main/java/com/box/sdk/BoxWebHook.java
stanislav-dvorscak: Fixed.

src/main/java/com/box/sdk/BoxWebHookSignatureVerifier.java
stanislav-dvorscak: fixed

src/main/java/com/box/sdk/BoxWebHook.java
stanislav-dvorscak: done

src/main/java/com/box/sdk/BoxWebHook.java
```diff
@@ -0,0 +1,687 @@
...
+     *
+     * @author Stanislav Dvorscak
+     *
+     */
+    public class Info extends BoxResource.Info {
+
+        /**
+         * @see #getTarget()
+         */
+        private Target target;
```
sekhridhiraj: Do we need separate class "Target"? Should we not use BoxResource here ?

src/main/java/com/box/sdk/BoxAPIRequest.java
stanislav-dvorscak: Yes, it can be a part of feedback process, sorry I am only hate String constants, where enum or codebook can be used. It was impulsive reaction, if you want String, I can revert it.

src/main/java/com/box/sdk/BoxWebHook.java
```diff
@@ -0,0 +1,687 @@
...
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public BoxWebHook getResource() {
+            return BoxWebHook.this;
+        }
+
+        /**
+         * @return WebHook target / {@link BoxItem}.
```
sekhridhiraj: Documentation: Not sure if we want to link to BoxItem.

src/main/java/com/box/sdk/BoxWebHook.java
stanislav-dvorscak: I have thought already about this, but it will be needed to alter all BoxResources to provide a method to get a type of resource, can I alter all resources?

src/main/java/com/box/sdk/BoxWebHookSignatureVerifier.java
```diff
@@ -0,0 +1,256 @@
...
+        try {
+            mac.init(new SecretKeySpec(key.getBytes(UTF_8), algorithm.javaProviderName));
+            mac.update(UTF_8.encode(webHookPayload));
+            mac.update(UTF_8.encode(deliveryTimestamp));
+        } catch (InvalidKeyException e) {
+            throw new IllegalArgumentException("Invalid key: ", e);
+        } finally {
+            MAC_POOL.release(mac);
+        }
+        return mac.doFinal();
```
sekhridhiraj: I guess mac.doFinal() should be performed before returning the Mac to the pool? Else it might lead to race condition, where some other thread re-uses this Mac (before we call mac.doFinal() in the current thread).

stanislav-dvorscak: Yes it is good catch, it was typo issue. All operations have to be between try - finally.

src/main/java/com/box/sdk/BoxWebHook.java
sekhridhiraj: @stanislav-dvorscak: For me adding a new method to BoxResource seems to be a better choice. This would be backward compatible and help get rid of Target and TargetType. However, I will let @gcurtis or @tarrencev confirm that.

src/main/java/com/box/sdk/BoxAPIRequest.java
gcurtis: Yeah, we can always come back to this later. I'd prefer if it was left as a String for now.

src/main/java/com/box/sdk/BoxWebHook.java
stanislav-dvorscak: I am still waiting for answer, confirmation for this, to alter box resource and classes which extends it. I think it will be good change, can I perform such change?

dmaynard: @stanislav-dvorscak   Yes   agreed.  You can add this change.

src/main/java/com/box/sdk/BoxResource.java
```diff
@@ -22,30 +35,66 @@ public BoxResource(BoxAPIConnection api, String id) {
...
+    }
+
+    /**
+     * Resolves {@link BoxResourceType} for a provided {@link BoxResource} {@link Class}.
+     *
+     * @param clazz
+     *            {@link BoxResource} type
+     * @return resolved {@link BoxResourceType#value()}
+     */
+    public static String getType(Class<? extends BoxResource> clazz) {
```
dmaynard: getType seems too generic.   I think getResourceType would make the code clearer.

src/main/java/com/box/sdk/BoxResourceIterable.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.NoSuchElementException;
+
+import com.eclipsesource.json.JsonArray;
+import com.eclipsesource.json.JsonObject;
+import com.eclipsesource.json.JsonValue;
+
+/**
+ * Common implementation for paging support.
+ *
+ * @author Stanislav Dvorscak
```
dmaynard: Please remove @author comments from all files.   (Box is the implied author)

src/main/java/com/box/sdk/BoxResource.java
```diff
@@ -1,5 +1,11 @@
 package com.box.sdk;
```
dmaynard: Not sure why but I am getting a checkstyle error on this page:
CheckStylePluginException: CheckStyle threw an exception when scanning: BoxResource.java: Couldn't create Checker from Box SDK CheckStyle.
Could we perhaps isolate isolate where it is happening and turn checkstyle off for that portion of the code?

src/main/java/com/box/sdk/BoxResource.java
stanislav-dvorscak: thanks, fixed

src/main/java/com/box/sdk/BoxResourceIterable.java
stanislav-dvorscak: Done, I have also updated my eclipse template to avoid it in the future.

src/main/java/com/box/sdk/BoxResource.java
stanislav-dvorscak: it has to relate to your local environment, which version of checkstyle do you have?

src/main/java/com/box/sdk/BoxResourceIterable.java
dmaynard: The Web Hook endpoint Get Webhooks is based  on a limit and marker scheme like a lot of the V2 endpoints.   While we agree that your new Iterator implementation is an elegant solution, we don't think it should be included as part of this pull request which should be focused only on providing a Java Implementation for the WebHooks API.  We do not want to one form for webhooks and another form for the other paged resources.   At some point point in the future (perhaps near future) we may want to add your implementation to all the other endpoints so please keep the code around.  However for this pull request please use the  JSONIterator interface.

src/main/java/com/box/sdk/BoxResourceIterable.java
```diff
@@ -0,0 +1,183 @@
...
+        return new IteratorImpl();
+    }
+
+    /**
+     * Paging implementation.
+     */
+    private class IteratorImpl implements Iterator<T> {
+
+        /**
+         * Base 64 encoded string that represents where the paging should being. It should be left blank to begin
```
dmaynard: typo? being -> begin

src/main/java/com/box/sdk/BoxResourceIterable.java
dmaynard: The Web Hook endpoint Get Webhooks is based on a limit and marker scheme like a lot of the V2 endpoints. While we agree that your new Iterator implementation is an elegant solution, we don't think it should be included as part of this pull request which should be focused only on providing a Java Implementation for the WebHooks API. We do not want to one form for webhooks and another form for the other paged resources. At some point point in the future (perhaps near future) we may want to add your implementation to all the other endpoints so please keep the code around. We agree you new implementation will be better for moving forward to Java 8. However for this pull request please use the JSONIterator interface.

src/main/java/com/box/sdk/BoxWebHookSignatureVerifier.java
```diff
@@ -0,0 +1,253 @@
+package com.box.sdk;
```
dmaynard: Question:  Does this new BoxWebHooksSignatureVerifier place any addition requirements on the Java SDK such as installing special cryptographic libraries?   Or were these special libraries already required for the  Box Java SDK?   If there are any new requirements they must be documented in the setup instructions for the SDK.

dmaynard:
@stanislav-dvorscak  could you please remove all the @author  comments.

We prefer to just have Box be the default author so that all feedback comes through our developer channels.  Also remove the @author Vladimir  comments.  
Thank you

src/main/java/com/box/sdk/BoxWebHookSignatureVerifier.java
stanislav-dvorscak: There is no additional third party library. We use "javax.crypto". Which is direct part of Java SDK. Potential issue can be with non Oracle JDK, e.g.: OpenJDK, but it uses only simple "HmacSHA256", which I think is supported by OpenJDK since 2003. I am not sure about other Java SDK implementations.

stanislav-dvorscak:
@dmaynard I have rechecked the source code, and all _@author_ comments were already removed. Diff from this discussions does not reflect recent state.

dmaynard:
Replaced by Pull Request #288

PR #295, opened 2016-09-28
Collections / Create or Delete

stanislav-dvorscak:


boxcla:
Verified that @stanislav-dvorscak has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/BoxItem.java
```diff
@@ -36,6 +36,11 @@ public BoxItem(BoxAPIConnection api, String id) {
     }
 
     /**
+     * @return base URL for current item
```
dmaynard: This name is confusing for two reasons.  One is the same name as a method on the API class, and two it isn't really returinging a base URL. It is returning a base URL plus an item specifier.   I suggest we change the name getItemURL and change the javadocs to reflect what the method really returns.

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -78,6 +78,14 @@ public BoxFile(BoxAPIConnection api, String id) {
         super(api, id);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected URL getBaseURL() {
```
dmaynard: See the suggestion on BoxItem.java to rename this get ItemURL

src/main/java/com/box/sdk/BoxFolder.java
```diff
@@ -61,6 +61,14 @@ public static BoxFolder getRootFolder(BoxAPIConnection api) {
     }
 
     /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected URL getBaseURL() {
```
dmaynard: See the suggestion on BoxItem.java to rename this get ItemURL

src/main/java/com/box/sdk/BoxItem.java
```diff
@@ -36,6 +36,11 @@ public BoxItem(BoxAPIConnection api, String id) {
...
+     */
+    protected abstract URL getBaseURL();
```
dmaynard: change to getItemURL

src/main/java/com/box/sdk/BoxWebLink.java
```diff
@@ -37,6 +37,14 @@ public BoxWebLink(BoxAPIConnection api, String id) {
         super(api, id);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected URL getBaseURL() {
```
dmaynard: See the suggestion on BoxItem.java to rename this get ItemURL

config/checkstyle/checkstyle.xml
```diff
@@ -41,7 +41,9 @@
 
     <!-- Checks whether files end with a new line.                        -->
     <!-- See http://checkstyle.sf.net/config_misc.html#NewlineAtEndOfFile -->
-    <module name="NewlineAtEndOfFile"/>
+    <module name="NewlineAtEndOfFile">
+		<property name="lineSeparator" value="lf"/>
```
gleb-urvanov: Checkstyle was changed exactly because Windows system is in use. Windows system default is "crlf", and all the existing code has "lf"- so without that change checkstyle on Windows fails on every file, expecting "crlf" at the end of file.

gleb-urvanov:
Requested changes are made now.

PR #297, opened 2016-09-28
Fix metadata filter methods (Closes #296)

### LINKED ISSUE #296: Metadata filters erroneously converting template key to lower case
Metadata templates are currently case sensitive - for example "keyName" and "keyname" do not point to the same template key instance. Currently, the BoxMetadataFilter.java methods addNumberRangeFilter() and addDateRangeFilter() incorrectly convert the keys to lower case when adding the filters to a search.

ppaulauskas:
Template keys are case sensitive, we should not be converting them to lower case when adding them to the search.


boxcla:
Verified that @ppaulauskas has signed the CLA. Thanks for the pull request!

PR #307, opened 2016-10-11
Create Metadata on Folder

gleb-urvanov:
Create Metadata on Folder feature implemented as createMetadata(...) methods of Metadata class.
createMetadata() methods of BoxFile class refactored to avoid code duplication.


boxcla:

PR #323, opened 2016-10-18
Fixed #316

### LINKED ISSUE #316: Get Memberships for Group: no result pagination
`BoxGroup.getMemberships()` doesn't return the correct result set if there are more than 1000 memberships for a group.

**The HTTP reference states the maximum per-page limit is 1000; but `BoxGroup.getMemberships()` doesn't specify it and doesn't paginate the results.**

dorfire:
To be merged with #321, some day.


boxcla:
Verified that @dorfire has signed the CLA. Thanks for the pull request!

gcurtis:
Thanks for the contribution! Unfortunately, we can't change the method signature until we do a new major release, so I've tweaked your change and submitted it as #336.

Issue #335, opened 2016-11-07
BoxGroup.getMemberships() should return an Iterable for pagination


gcurtis:
Right now it returns a `Collection<BoxGroupMembership.Info>` which prevents us from being able to automatically paginate results.

Related to issue #316 and PR #323.

mattwiller:
I'm closing this since the immediate issue is resolved, but we'll add to our backlog the cleanup work to return the iterator directly in the next major version.

PR #336, opened 2016-11-07
Return all results for BoxGroup.getMemberships()

### LINKED ISSUE #316: Get Memberships for Group: no result pagination

gcurtis:
This method wasn't correctly paginating results and only returned the first 100 memberships. This change returns all of them by automatically paging.

Ideally, we'd like to return an Iterator that allows for lazy paging, but that would require a method signature change. This is tracked as a breaking change in #335.

This change was originally authored by Dor Fire, but modified to prevent the backwards-incompatible change.

Fixes #316.


boxcla:

PR #340, opened 2016-11-09
Fix for BoxTaskAssignment

gleb-urvanov:
Wrong URLTemplate of BoxTaskAssignment value altered to match API reference.
BoxTaskAssignment.update method signature was altered to accept BoxTaskAssignment.Info object as a param instead of BoxTask.Info. So that change can cause some compatibility issues, if any third party code uses that method with wrong signature.


boxcla:

PR #343, opened 2016-11-10
Add -- to last boundary

### LINKED ISSUE #334: Multipart Request needs ending "--" when going through Layer7
When sending an upload a file to Box, I need to go through Layer7 as part of our policies. Layer7 expects a "--" after the last boundary. After looking through other self generate multipart request, the "--" after the last boundary seems like an accepted behavior. Adding this will not break the request when going to Box.

Kalrock:
Added the -- to the last boundary to fix and issue when going through
Layer7.

Closes #334.

### DIFF
--- a/src/main/java/com/box/sdk/BoxMultipartRequest.java
+++ b/src/main/java/com/box/sdk/BoxMultipartRequest.java
@@ -140,6 +140,7 @@ public class BoxMultipartRequest extends BoxAPIRequest {
             }
 
             this.writeBoundary();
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @Kalrock, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

gcurtis:
Thanks! This will be going out in our next v2.2.0 release, which should be next Friday.

PR #357, opened 2016-11-11
Ew 232

dmaynard:
Added some missing Event types, and updated the documents.  (also tested the API and found one back end bug (TASK_CREATE is a Enterprise event where as COMMENT_CREATE is a User Event).
Also identified some missing TASKS in the API document  (TASK_ASSIGN_COMPLETE)

### DIFF
--- a/doc/events.md
+++ b/doc/events.md
@@ -1,7 +1,7 @@
-Events
-======
+User Events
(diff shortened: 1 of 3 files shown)


boxcla:

PR #365, opened 2016-11-14
Get Metadata Template and Get Enterprise Metadata features

gleb-urvanov:
"Get Metadata Template" and "Get Enterprise Metadata" features implemented as functions of introduced MetadataTemplate object. Functions were documented and covered with unit tests.


boxcla:

PR #370, opened 2016-11-16
Fixes and unit tests for Tasks

gleb-urvanov:
Added "fields" param support for Get Task, Get Task Assignment and Get Assignments features;

Added paging support for Get Assignments;

Added opportunity to create Task Assignment [using user's login](https://docs.box.com/reference#create-a-task-assignment);

Methods were covered with unit tests.


boxcla:

PR #374, opened 2016-11-18
Fixed and tests for Delete and Promote File Version

gleb-urvanov:
BoxFileVersion object refactored:
- provided opportunity to set associated file's id not only thru constructor;
-  API response json object, obtained thru promote() call, now parsed to update fields of the object;

Unit tests were added for Delete Old Version and Promote Version (delete() and promote() methods)


boxcla:

PR #385, opened 2017-01-27
Some changes to enable getting log output from Integration Tests

dmaynard:


boxcla:

src/test/java/com/box/sdk/TestConfig.java
```diff
@@ -25,7 +25,7 @@
 
     private TestConfig() { }
 
-    public static void setLogLevel(String levelString) {
+    public static Logger  setLogLevel(String levelString) {
```
gcurtis: What about changing the name to `enableLogger(String)`? I think that's a slightly better name for a function that actually returns a logger.

src/test/java/com/box/sdk/EventStreamTest.java
```diff
@@ -28,8 +28,8 @@
     @Test
     @Category(IntegrationTest.class)
     public void receiveEventsForFolderCreateAndFolderDelete() throws InterruptedException {
+        // Logger logger = TestConfig.setLogLevel("FINE");
```
gcurtis: Forgot to remove this.

PR #423, opened 2017-04-26
multiput uploads for large files using parallel http connections for …

iamharish:
…supercharged performance (when network bandwidth allows)


boxcla:

Issue #429, opened 2017-05-04
Should not allow 2+ As-User headers


nddipiazza:
Should create an error or just ignore when a request tries to add 2+ `As-User` headers to a Box api request.

Because often times when I'm trying to re-use api connections I use the setRequestInterceptor and `headers.add` to the headers but I have no way to know if I already added the header yet. 

I can't prove it but I don't really think my code is actually adding 2 as-user headers... it might be something to do with how the set request interceptor method works. But either way, would be nice to just prevent the possibility of that.

iamharish:
Fixed in https://github.com/box/box-java-sdk/pull/450

PR #431, opened 2017-05-04
Don't allow multiple As-User headers in Box api requests

nddipiazza:
Don't allow people to create duplicate as-user headers.

This can prevent issues when people use setRequestInterceptor to simulate users but accidentally end up with multiple as-users when sharing api connections.

https://github.com/box/box-java-sdk/issues/429

this probably isn't the best way to do it... but it was the easiest for me to prevent people from accidentally getting multiple as-users from the request interceptor stuff.

Failure to do this will result in error message like below if you have more than one as-user specified:

`Cannot make request On Behalf of User is not found "235235233, 2352352352"`

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIRequest.java
+++ b/src/main/java/com/box/sdk/BoxAPIRequest.java
@@ -9,6 +9,7 @@ import java.net.MalformedURLException;
 import java.net.ProtocolException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @nddipiazza, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

nddipiazza:
Done.

nddipiazza:
CLA signed

boxcla:
Verified that @nddipiazza has just signed the CLA. Thanks, and we look forward to your contribution.

iamharish:
@nddipiazza Thanks for submitting a pull request. It looks good. I just used your logic but made it look, a bit, simple. Can you submit a new pull request please. 

```public void addHeader(String key, String value) {
        if ("As-User".equals(key)) {
            int index = this.headers.indexOf("As-User");
            if ( index > -1) {
                this.headers.remove(index);
            }
         }
        this.headers.add(new RequestHeader(key, value));
    }
```

If not, I can fix it in next pull request as well.

iamharish:
Resolved this issue in a different commit

PR #432, opened 2017-05-06
Issue #438 and #431 - For API 403 rate limit responses, use the `RetryAfter` header in the retry logic.

nddipiazza:
Addresses https://github.com/box/box-java-sdk/issues/438 The box java sdk needs to have api requests have the ability to honor the RetryAfter header if 403 error code is returned.

When you get a 403 with a `RetryAfter` header it should display a message like this: `Encountered a 403 Error with Retry-After header of 1s. Sleeping this thread that amount and trying again.` then it should sleep the current thread until the time has elapsed before retrying.

I also added in https://github.com/box/box-java-sdk/pull/431

And I added an API count ticker on the FINE logging level for people who need to keep track of exact counts when doing large numbers of api calls during recursive directory file access

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ sourceCompatibility = 1.6
 
 group = 'com.box'
 archivesBaseName = 'box-java-sdk'
-version = '2.4.0-SNAPSHOT'
+version = '2.4.0'
 
 repositories {
     mavenCentral()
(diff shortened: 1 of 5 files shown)


boxcla:
Verified that @nddipiazza has signed the CLA. Thanks for the pull request!

mattwiller:
Closing since we're not planning to use this retry strategy — see #438 for more details.

PR #437, opened 2017-05-23
Token exchange

iamharish:
https://developer.box.com/docs/getting-started-with-new-box-view#section-retrieve-a-file-token-using-token-exchange-optional

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIConnection.java
+++ b/src/main/java/com/box/sdk/BoxAPIConnection.java
@@ -607,6 +607,57 @@ public class BoxAPIConnection {
     }
 
     /**
+     * Get a lower-scoped token restricted to a resource for the list of scopes that are passed.
+     * @param scopes the list of scopes to which the new token should be restricted for
+     * @param resource the resource for which the new token has to be obtained
+     * @return scopedToken which has access token and other details
+     */
+    public ScopedToken getLowerScopedToken(List<String> scopes, String resource) {
+        assert (scopes != null);
+        assert (scopes.size() > 0);
+        assert (resource != null);
+        URL url = null;
+        try {
+            url = new URL(this.getTokenURL());
+        } catch (MalformedURLException e) {
+            assert false : "An invalid refresh URL indicates a bug in the SDK.";
+            throw new RuntimeException("An invalid refresh URL indicates a bug in the SDK.", e);
+        }
+        StringBuilder spaceSeparatedScopes = new StringBuilder();
+
+        for (String scope : scopes) {
+            spaceSeparatedScopes.append(scope);
+            spaceSeparatedScopes.append(" ");
+        }
+        //Delete the extra space at the end
+        spaceSeparatedScopes.deleteCharAt(spaceSeparatedScopes.length() - 1);
+
+        //this.getAccessToken() ensures we have a valid access token
+        String urlParameters = String.format("grant_type=urn:ietf:params:oauth:grant-type:token-exchange"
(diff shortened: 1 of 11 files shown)


boxcla:

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -607,6 +607,57 @@ public void setRequestInterceptor(RequestInterceptor interceptor) {
...
+            BoxJSONResponse response = (BoxJSONResponse) request.send();
+            json = response.getJSON();
+        } catch (BoxAPIException e) {
+            this.notifyError(e);
+            throw e;
+        }
+
+        JsonObject jsonObject = JsonObject.readFrom(json);
+        ScopedToken token = new ScopedToken(jsonObject);
+        token.setObtainedAt(System.currentTimeMillis());
```
KarthikShanmugasundaram: Ideally it should be the actual token creation time. As people may get some gaps in the time in milli seconds. But at this moment we can't do anything else as the expiration could be configurable.

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -607,6 +607,57 @@ public void setRequestInterceptor(RequestInterceptor interceptor) {
...
+            throw new RuntimeException("An invalid refresh URL indicates a bug in the SDK.", e);
+        }
+        StringBuilder spaceSeparatedScopes = new StringBuilder();
+
+        for (String scope : scopes) {
+            spaceSeparatedScopes.append(scope);
+            spaceSeparatedScopes.append(" ");
+        }
+        //Delete the extra space at the end
+        spaceSeparatedScopes.deleteCharAt(spaceSeparatedScopes.length() - 1);
```
iamharish: I'm using a StringBuilder and not a String. It does not copy but modifies the current StringBuilder object in memory. So should not be a costly operation

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -607,6 +607,57 @@ public void setRequestInterceptor(RequestInterceptor interceptor) {
     }
 
     /**
+     * Get a lower-scoped token restricted to a resource for the list of scopes that are passed.
+     * @param scopes the list of scopes to which the new token should be restricted for
+     * @param resource the resource for which the new token has to be obtained
+     * @return scopedToken which has access token and other details
+     */
+    public ScopedToken getLowerScopedToken(List<String> scopes, String resource) {
+        assert (scopes != null);
```
iamharish: It is

src/main/java/com/box/sdk/BoxAPIConnection.java
iamharish: Since we don't have the issued time, sadly, this gap is something we have to deal with

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -607,6 +607,57 @@ public void setRequestInterceptor(RequestInterceptor interceptor) {
     }
 
     /**
+     * Get a lower-scoped token restricted to a resource for the list of scopes that are passed.
+     * @param scopes the list of scopes to which the new token should be restricted for
+     * @param resource the resource for which the new token has to be obtained
+     * @return scopedToken which has access token and other details
+     */
+    public ScopedToken getLowerScopedToken(List<String> scopes, String resource) {
```
coolboy: .net sdk is accepting full url

PR #441, opened 2017-06-05
Issue433

### LINKED ISSUE #433: Confusion about inconsistency of BoxSearch.java document and API references
In file BoxSearch.java, line 31, the doc says 1000 as max number.
But in https://developer.box.com/v2.0/reference#searching-for-content, the max number of limit is 200.

I know Java SDK is also using "https://api.box.com/2.0/". 

So which one is correct reference?

iamharish:
fixes #433


boxcla:

PR #465, opened 2017-09-02
Representations api

iamharish:
- Utility method to pass **x-rep-hints** header as part of getting representations
- Representations object has properties or metadata as JsonObject's instead of defined classes as per request from API team for being generic enough to handle future changes


boxcla:

PR #467, opened 2017-09-13
1. additional constructors using box config, similar to additional ge…

iamharish:
…tAppEnterpriseConnection methods that use the box config

2. Example using one of the newly added constructors


boxcla:

PR #474, opened 2017-10-05
Additional test in response for Batch Api

carycheng77:


boxcla:
Verified that @carycheng77 has signed the CLA. Thanks for the pull request!

Issue #387, opened 2017-01-27
Request support for tracking_codes in the BoxUser.Info class


paapfly:
Need support for tracking_codes in the BoxUser.Info class

PR #476, opened 2017-10-19
Support traking_codes on BoxUser.Info 

silvestrelosada:
Current pull request solve the issue #387. Tracking codes are pairs of key/value elements. It is returned as a Map. It can also be obtained as List<String>, but I find better have them as Map. If you prefer as List let me know I can implement it. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,11 @@ build/
 .DS_Store
 src/test/config/config.properties
 /target/
+.settings/
 .idea/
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @silvestrelosada, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

PR #480, opened 2017-11-09
Avoid502retry

jtrujillo3:
The reason for this request comes out of a need for moveFolderToUser in BoxUser.java to not retry when receiving a 502 response. This is because a 502 response could come back but the action performed in the background and completed. I have created duplicate methods in BoxApiRequest and did not alter existing methods.  These duplicate methods are only called when moveFolderToUser is called.

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIRequest.java
+++ b/src/main/java/com/box/sdk/BoxAPIRequest.java
@@ -240,37 +240,36 @@ public class BoxAPIRequest {
         return this.send(null);
     }
 
-	/**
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @jtrujillo3 has signed the CLA. Thanks for the pull request!

mattwiller:
This seems like a pretty complicated solution to a very specific problem.  In general, I'm not sure this is the right approach to the issues with this endpoint.  I think the 502 issue needs to be resolved with the API, so I'm going to go ahead and close this PR.

PR #486, opened 2017-11-21
Support traking_codes on BoxUser.Info

silvestrelosada:
Current pull request solve the issue #387. Tracking codes are pairs of key/value elements. It is returned as a Map. It can also be obtained as List, but I find better have them as Map. If you prefer as List let me know I can implement it. Current request fix check style


boxcla:
Verified that @silvestrelosada has signed the CLA. Thanks for the pull request!

PR #487, opened 2017-11-27
Add support for User Tracking Codes

silvestrelosada:
Current pull request solve the issue #387. Tracking codes are pairs of key/value elements. It is returned as a Map. It can also be obtained as List, but I find better have them as Map. If you prefer as List let me know I can implement it. Current request fix check style


boxcla:
Verified that @silvestrelosada has signed the CLA. Thanks for the pull request!

mattwiller:
@silvestrelosada Would you be able to add some tests for this functionality?  Ideally we would have a unit test using WireMock (see [this example test][example-test]) to return some stub data (taken from the real API but scrubbed of personal data) and verify that this logic actually parses that data correctly.  I'm happy to help if you can't get to it or need some assistance — just let me know!

[example-test]: https://github.com/box/box-java-sdk/blob/master/src/test/java/com/box/sdk/BoxUserTest.java#L55

silvestrelosada:
One question tracking codes are specific functionality that has to be
enabled on box, will the tracking codes be available on wire mock?

In any case I'll try to o do it in case I found some problems I let you
know.

Best

mattwiller:
@silvestrelosada Wiremock just returns stub data, and doesn't actually hit the Box API, so you can definitely set up a unit test that returns tracking codes without needing to do any setup with a real Box account.

silvestrelosada:
Thanks a lot,

Let me try to prepare a test during next week, really busy those days

Issue #493, opened 2017-12-18
No support for indefinite file locking


meavenson:
Currently, there appears to be no support for indefinitely locking a Box file, as the two existing lock methods in BoxFile require a non-null expiry date to be supplied.  If you attempt to pass null instead of a Date object, a null pointer exception is thrown.

carycheng77:
Hey @meavenson, thank you for letting us know, we'll get started on that fix!

meavenson:
Thanks @carycheng77 !  I submitted a pull request for review, in case you'd like to use my fix.

carycheng77:
#494

PR #494, opened 2017-12-18
Adding support for indefinite Box file locking by allowing null expir…

meavenson:
…y dates.  #493

### DIFF
--- a/src/test/java/com/box/sdk/BoxFileTest.java
+++ b/src/test/java/com/box/sdk/BoxFileTest.java
@@ -644,14 +644,14 @@ public class BoxFileTest {
         BoxFile.Info updatedFileInfo = uploadedFile.getInfo("lock");
         assertThat(updatedFileInfo.getLock(), is(nullValue()));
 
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @meavenson has signed the CLA. Thanks for the pull request!

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -644,6 +644,30 @@ public void fileLockAndUnlockSucceeds() {
         BoxFile.Info updatedFileInfo = uploadedFile.getInfo("lock");
         assertThat(updatedFileInfo.getLock(), is(nullValue()));
 
+        uploadedFile.lock(false);
```
carycheng77: Hey @meavenson, thank you so much for the pull request! 

This looks great, the only thing is it would be great to assert that getIsDownloadPrevented() asserts to true here as it is false by default and this way we can ensure that the out of the ordinary case passes. Otherwise everything else looks great and we can merge this in today.  

Thank you!

PR #497, opened 2017-12-21
Add missing webhook triggers to enum

### LINKED ISSUE #491: Webhook Issue - METADATA_INSTANCE.UPDATED is not on the list of triggers
It seems there is an issue in the Box Java SDK on which our production application was built. In particular, we attempt to set up a webhook on a folder and pass in a METADATA_INSTANCE.UPDATED trigger. The webhook does get created, however, when we inspect this webhook (using the native API or the Box CLI) METADATA_INSTANCE.UPDATED is not on the list of triggers.

We inspected the Java SDK source code on github and observed the trigger enumeration in the webhook class didn't have any METADATA triggers--this could explain what we have noticed.

If this is the case, is this a defect in the JDK? Can we request this get fixed as soon as possible? We have a production rollout coming up where this is crucial.

Thank you,

Ben Allen

mattwiller:
Fixes #491


boxcla:

PR #498, opened 2017-12-21
Fix MetadataTemplate.updateMetadataTemplate() type error

### LINKED ISSUE #489: BoxJSONRequest issue trying arrays in body like for template updates
When doing update on templates fails, since you are passing ArrayJSON as string to the BoxJSONRequest, but this class only works if you are passing an Object not an array, will fail on 👍 

```
public static MetadataTemplate updateMetadataTemplate(BoxAPIConnection api, String scope, String template,
            List<FieldOperation> fieldOperations) {

......

JsonArray array = new JsonArray();

        for (FieldOperation fieldOperation : fieldOperations) {
            JsonObject jsonObject = getFieldOperationJsonObject(fieldOperation);
            array.add(jsonObject);
        }

        QueryStringBuilder builder = new QueryStringBuilder();
        URL url = METADATA_TEMPLATE_URL_TEMPLATE.build(api.getBaseURL(), scope, template);
        BoxJSONRequest request = new BoxJSONRequest(api, url, "PUT");
        request.setBody(array.toString());


BoxJSONRequest class

 @Override
    public void setBody(String body) {
        super.setBody(body);
        this.jsonObject = JsonObject.readFrom(body);
    }
```

This fails with an exception. You need to consider that JsonObject.readFrom(body); wont work with Arrays, only objects.

mattwiller:
Loosen internal restrictions in BoxJSONRequest to allow for JSON
types other than objects, since metadata template updates use a
JSON array as the top level body.

Fixes #489

### DIFF
--- a/src/main/java/com/box/sdk/BatchAPIRequest.java
+++ b/src/main/java/com/box/sdk/BatchAPIRequest.java
@@ -61,7 +61,7 @@ public class BatchAPIRequest extends BoxJSONRequest {
             //If the actual request has a JSON body then add it to vatch request
             if (request instanceof BoxJSONRequest) {
                 BoxJSONRequest jsonRequest = (BoxJSONRequest) request;
-                batchRequest.add("body", jsonRequest.getBodyAsJsonObject());
+                batchRequest.add("body", jsonRequest.getBodyAsJsonValue());
             }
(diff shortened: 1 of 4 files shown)


boxcla:

PR #499, opened 2017-12-21
Add optional is_confirmed parameter for adding user email alias

### LINKED ISSUE #469: Why addEmailAlias does not take in input is_confirmed parameter?
Hi,

in order to add an email alias, domain should be allowed into enterprise settings.
But to bypass this check `email_aliases` endpoint provides also an `is_confirmed` parameter.
I'm referring to [this forum thread](https://community.box.com/t5/Box-Developer-Forum/Add-Email-Alias-API-returns-bad-request/td-p/19822).

Why there isn't such parameter into [addEmailAlias](https://github.com/box/box-java-sdk/blob/v2.8.1/src/main/java/com/box/sdk/BoxUser.java#L339) method?

mattwiller:
Fixes #469


boxcla:

src/main/java/com/box/sdk/BoxUser.java
```diff
@@ -337,11 +337,27 @@ public static BoxUser getCurrentUser(BoxAPIConnection api) {
...
+     */
+    public EmailAlias addEmailAlias(String email, boolean isConfirmed) {
         URL url = EMAIL_ALIASES_URL_TEMPLATE.build(this.getAPI().getBaseURL(), this.getID());
         BoxJSONRequest request = new BoxJSONRequest(this.getAPI(), url, "POST");
 
         JsonObject requestJSON = new JsonObject()
-            .add("email", email);
+                .add("email", email);
+
+        if (isConfirmed) {
```
iamharish: How is this different from the addEmailAlias method when isConfirmed is false. Also, I can't find any documentation about it in https://developer.box.com/reference#get-all-email-aliases-for-a-user

mattwiller: The `is_confirmed` parameter allows an admin to add an email alias for a user without needing the user to manually verify the email address.  I'll follow up about how we want to document it, but it's definitely something that's in use today.

iamharish: Since is_confirmed defaults to *false* in the `addEmailAlias(String)` method, can we just make the second method simple, something like `addConfirmedEmailAlias(String)` which internally defaults the is_confirmed param as *true*.

mattwiller: Is there a way to do that without duplicating logic?  I'm looking at the code and it seems like in order to have two simple methods like that I would need to copy the main logic for the method across both.

iamharish:
```
public EmailAlias addEmailAlias(String email)
public EmailAlias addConfirmedEmailAlias(String email)
private EmailAlias addEmailAlias(String email, boolean isConfirmed)
```
Both the public methods can call this private method

PR #502, opened 2017-12-21
Add option to pass file SHA-1 hash for upload integrity

### LINKED ISSUE #328: Can't set the 'Content-MD5' header when uploading a file
Example from the deprecated SDK:
`uploadRequestObject.setContentMD5(sha1);`

mattwiller:
Added a sha1 attribute to FileUploadParams to enable sending the
expected hash of the file for verification that the file was not
corrupted in transit.

Fixes #328

### DIFF
--- a/src/main/java/com/box/sdk/BoxFolder.java
+++ b/src/main/java/com/box/sdk/BoxFolder.java
@@ -493,7 +493,7 @@ public class BoxFolder extends BoxItem implements Iterable<BoxItem.Info> {
             fieldJSON.add("content_modified_at", BoxDateFormat.format(uploadParams.getModified()));
         }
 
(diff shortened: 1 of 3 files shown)


boxcla:

src/main/java/com/box/sdk/BoxFolder.java
```diff
@@ -493,6 +493,10 @@ public void canUpload(String name, long fileSize) {
             fieldJSON.add("content_modified_at", BoxDateFormat.format(uploadParams.getModified()));
         }
 
+        if (uploadParams.getSHA1() != null) {
```
iamharish: Better to check that string is not empty
uploadParams.getSHA1() != null && !uploadParams.getSHA1().isEmpty()

Issue #61, opened 2015-01-22
Invite existing user to join an enterprise


gcurtis:
Implement the `/invites` endpoint.

PR #504, opened 2017-12-21
Implement invite user to enterprise

mattwiller:


boxcla:

mattwiller:
Fixes #61

src/test/java/BoxInviteTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+/**
+ * {@link BoxInvite} related tests.
+ */
+public class BoxInviteTest {
```
iamharish: Would you be able to add an integration test as this is a new api?

mattwiller:
@iamharish This would be kind of tricky to add integration tests for, since it involves setting up a user account that isn't in the enterprise.  It's not easy to create that through the API today — certainly not in a way that would be easily repeatable.  Do you have any ideas on how we could set up an end-to-end test for this?

PR #505, opened 2017-12-24
Change wiremock port

### LINKED ISSUE #60: Pick a different port for web server in unit tests
WireMock runs on port 8080, which is a pretty common port for running local web servers, so it may already be occupied. People will be less likely to encounter errors when running unit tests if a more random port were used.

mattwiller:
Changes wiremock in tests to use a private port instead of 8080, which is popular with web servers.

Fixes #60


boxcla:

PR #506, opened 2017-12-28
Search readme update

carycheng:


doc/search.md
```diff
@@ -4,3 +4,56 @@ Search
 Different examples of search which have been implemented in the example [SearchExamplesAsAppUser.java](https://github.com/box/box-java-sdk/blob/master/src/example/java/com/box/sdk/example/SearchExamplesAsAppUser.java)
 
 For more information refer [search documentation](https://developer.box.com/v2.0/reference#searching-for-content)
+
+* [Search with Offset](#search-with-offset)
+* [Search with Limit](#search-with-limit)
+* [Search with Parameters](#search-with-parameters)
+
+Search with Offset
```
mattwiller: Since these all look like they refer to the same method, maybe combine into a single section that explains all three params?

carycheng77: @mattwiller, good point, will do!

doc/search.md
```diff
@@ -4,3 +4,37 @@ Search
 Different examples of search which have been implemented in the example [SearchExamplesAsAppUser.java](https://github.com/box/box-java-sdk/blob/master/src/example/java/com/box/sdk/example/SearchExamplesAsAppUser.java)
 
 For more information refer [search documentation](https://developer.box.com/v2.0/reference#searching-for-content)
+
+* [Search with Offset](#search-with-offset)
```
mattwiller: This needs to be updated to reflect the new section heading below

PR #507, opened 2018-01-02
Events log limit

carycheng77:
Adds a limit field to getEnterpriseEvents()

Thanks Matt!


boxcla:
Verified that @carycheng77 has signed the CLA. Thanks for the pull request!

Issue #338, opened 2016-11-08
Remove the deprecated BoxFolder.search() method


gcurtis:
It has been deprecated in favor of the `BoxSearch` class.

mattwiller:
I'm going to close this issue, since we're definitely going to remove any deprecated methods with the next major version release but we don't have a timeline for when such a release might be made and so this isn't really useful to track here.

PR #509, opened 2018-01-06
remove boxfolder search because deprecated

carycheng:
@mattwiller, not sure the best way to remove this from issue #338. 

This method was already marked @deprecated and this endpoint probably doesn't return what the user expects anymore but is it okay to remove it like we are doing here? Let me know, thanks!


boxcla:

PR #510, opened 2018-01-09
Add support for token revocation

### LINKED ISSUE #405: Add a way to revoke OAuth tokens
Best I can tell, this is not a supported feature of the java sdk, so we have to git the revoke endpoint directly.

Reference:
https://docs.box.com/v2.0/docs/oauth-20#section-7-destroying-tokens

Is that correct?

mattwiller:
Fixes #405


boxcla:

PR #524, opened 2018-01-17
Add updated file version upload endpoint

### LINKED ISSUE #518: BoxFile.uploadVersion returns void
When uploading a new version of a file, the BoxFile.uploadVersion method returns `void` while the other methods for uploading (`BoxFolder.uploadLargeFile`, `BoxFile.uploadLargeFile`, `BoxFolder.uploadFile`) all return type `BoxFile.Info`. It may be convenient for consistency to have an `uploadVersion` method that returns `BoxFile.Info`.

mattwiller:
Add a new BoxFile#uploadNewVersion() method that returns the
BoxFile.Info from the API response, and deprecate the old
BoxFile#uploadVersion() method that returns void.

Fixes #518


boxcla:

Issue #530, opened 2018-01-24
Large File Upload fails if file size is > 2,147,483,647


pmatte1:
Because the _processed_ variable in the uploadParts method of the LargeUploadFile is declared as an int it is maxed at  2,147,483,647.  However, the upload should handle files much larger than this, and therefore this creates a bug condition once the byte size exceeds  2,147,483,647.

The processed variable needs to be declared as a long, as it was done for the offset and filesize variables within that method.

![image](https://user-images.githubusercontent.com/35776253/35347356-e9d4b274-0102-11e8-9883-2e8c18c919e3.png)

mattwiller:
The fix for this issue has been released in v2.11.0 of the SDK

PR #531, opened 2018-01-24
Fix chunked upload for files > 2GB

mattwiller:


boxcla:

mattwiller:
Fixes #530

Issue #541, opened 2018-01-30
java.lang.OutOfMemory on Large File Upload


pmatte1:
When doing a large file upload on extremely large files (example 3gb file), the way uploadParts is implemented allows for the opportunity where too many LargeFileUploadTask objects, which hold the upload bytes, overrun the available Heap space.  This is because the ThreadPoolExecutor will queue up additional tasks, and with large files, the size of each part is large, and the time to execute the upload of that part is much longer.

Currently the only control in place is a 1 second delay if all current pool workers are occupied, which is not enough for these large scale documents.

**EXAMPLE**
I am testing with a 3gb document, where each part is 8.3mb, and the average upload time is >1 minute.  With the current logic, there are hundreds of queued LargeFileUploadTask objects, because they are only closing out 1 every minute.

This is made worse under a threaded condition, as the threads will share the available memory.

**RECOMMENDED SOLUTION:**
The simplest solution is to only read new bytes, and construct and queue the LargeFileUploadTask if the current Queue length is less than a certain value.  The recommended size of the queue is 1.5 times the number of processing threads.

That will allow the developer to control the memory consumption based upon the number of parallel threads being configured for execution in the LargeFileUpload class.

mattwiller:
Fixed by #543

PR #543, opened 2018-01-30
ISSUE #541 Large File Upload Memory

pmatte1:
Issue 541 - Out of Memory exceptions due to back log of queued items in the ThreadPoolExecutor

Resolution:
Added Queue Size Handling For ThreadPoolExecutor in LargeFileUpload.uploadParts().  This consisted of a line to determine the Max Queue Size, which was set to 1.5 times the max number of worker threads.  Then adding a check to see if the queue is less than that Max size, before reading in the next set of bytes and adding another task to the queue.

### DIFF
--- a/src/main/java/com/box/sdk/LargeFileUpload.java
+++ b/src/main/java/com/box/sdk/LargeFileUpload.java
@@ -166,7 +166,7 @@ public final class LargeFileUpload {
         long processed = 0;
         int partPostion = 0;
         //Set the Max Queue Size to 1.5x the number of processors
-        double maxQueueSizeDouble = Math.ceil(this.executorService.getMaximumPoolSize()*1.5);
+        double maxQueueSizeDouble = Math.ceil(this.executorService.getMaximumPoolSize() * 1.5);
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @pmatte1 has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/LargeFileUpload.java
```diff
@@ -165,6 +165,8 @@ public LargeFileUpload() {
         long offset = 0;
         long processed = 0;
         int partPostion = 0;
+        //Set the Max Queue Size to 1.5x the number of processors
+        int maxQueueSize = Double.valueOf(this.executorService.getMaximumPoolSize() * 1.5).intValue();
```
mattwiller: Could we do `Math.ceil()` or `Math.floor()` here for clarity?

pmatte1: Added the explicit call to Math.ceil() to explicitly round up.

PR #545, opened 2018-02-01
Fix collaboration notify parameter

### LINKED ISSUE #544: Issue with 'notify' parameter of collaborate method
Hi there,

I am trying to turn off email notification for collaborate invite,
but the invited person still receives email notification.

Below is the code I am using to do this:

```
BoxFolder testFolder = new BoxFolder(api, "FolderID");
testFolder.collaborate("Email Address", BoxCollaboration.Role.EDITOR, false, false);
```

Although I am setting the 3rd parameter ('notify') false,
the invited person still receives an email.

This seems like a glitch to me. Would you be able to fix this?
Thanks!

mattwiller:
Fixed a bug where the `notify` parameter when creating a collab
was placed in a header instead of the query string.

Also did some opportunistic refactoring to consolidate collab
creation API call logic in one place in the BoxCollaboration
class to avoid duplicate code.

Fixes #544

### DIFF
--- a/src/main/java/com/box/sdk/BoxCollaboration.java
+++ b/src/main/java/com/box/sdk/BoxCollaboration.java
@@ -47,6 +47,48 @@ public class BoxCollaboration extends BoxResource {
         super(api, id);
     }
 
+    /**
+     * Create a new collaboration object.
+     * @param api          the API connection used to make the request.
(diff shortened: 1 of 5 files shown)


boxcla:

PR #564, opened 2018-02-27
Delay successive API calls by EventStream

mattwiller:
Inserting a delay between polling calls to get more events in order to control the number of API calls made by the Event Stream.  Because there may be duplicates in the stream, it is currently possible for the stream to make 3-5 calls per second even in cases when the user sees no new events — this is obviously undesirable.  The default is now to wait 1s between calls, and is user-configurable.


boxcla:

PR #572, opened 2018-03-13
Adjust JWT expiration interval

mattwiller:
Setting the JWT expiration interval to the maximum allowed time of
60 seconds is not advisable, since the Box API server will actually
reject a JWT assertion if the exp claim is too far in the future.
Clock drift between the SDK and API of even a few seconds could
then cause the JWT to be rejected.  Setting the expiration
interval to 30 seconds instead should maximum the amount of clock
drift allowable before a JWT would get rejected, reducing the
chances of an error.


boxcla:

PR #595, opened 2018-04-27
Quick fix for NPE on request.toString() in interceptor

### LINKED ISSUE #133: NPE in RequestInterceptor.onRequest() BoxAPIRequest.toString()
We implemented a simple RequestInterceptor to log outbound BoxAPIRequest objects. We call the toString() method on the request object. The SDK (v 1.0.0) throws an NPE (null pointer) at:

``` java
for (Map.Entry<String, List<String>> entry : this.requestProperties.entrySet()) { ... }
```

because this.requestProperties is null when the interceptor is invoked. This is because the interceptor callback is being called before the requestProperties have been set. A fix to move the interceptor invocation to after line # 372 where this.requestProperties is initialized works.

``` java
this.requestProperties = connection.getRequestProperties();
```

(move interceptor calling to around line # 374)

``` java
if (this.api != null) {
    RequestInterceptor interceptor = this.api.getRequestInterceptor();
    if (interceptor != null) {
        BoxAPIResponse response = interceptor.onRequest(this);
        if (response != null) {
            return response;
        }
    }
}
```

mattwiller:
Small fix to prevent BoxAPIRequest#toString() from throwing a null
pointer exception inside of a request interceptor.  Ideally, the
request headers would always be available on the request object
instead of being partially stored on the connection, but in lieu
of that more involved fix, this should restore at least some
functionality in the meantime.

Fixes #133 (at least partially)

### DIFF
--- a/src/main/java/com/box/sdk/BoxAPIRequest.java
+++ b/src/main/java/com/box/sdk/BoxAPIRequest.java
@@ -361,27 +361,30 @@ public class BoxAPIRequest {
         builder.append(this.url.toString());
         builder.append(lineSeparator);
 
-        for (Map.Entry<String, List<String>> entry : this.requestProperties.entrySet()) {
-            List<String> nonEmptyValues = new ArrayList<String>();
-            for (String value : entry.getValue()) {
-                if (value != null && value.trim().length() != 0) {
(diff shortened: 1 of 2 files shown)


boxcla:

PR #597, opened 2018-05-01
Add support for multiselect fields in metadata

### LINKED ISSUE #592: Updating Multi-Select fields is currently not supported by the Box SDK
- [x] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
I have run into some issues using the Box SDK to update the metadata of fields that are new multi-select values. I can retrieve them easily enough.
I can not seem to find a feature on the Metadata object that allows you to set the field as directed in the link below. I have looked through the documentation and there are no examples I could find. 

There seems to be no way to set the value of a path to a JsonArray (or array of strings) on the Metadata class. I have upgraded my Java version of the SDK to 2.17.0

### Steps to Reproduce
This is best done with a unit test.
Create a metadata template on box with a multiselect field.
Assign the metadata to a file on box.
Create a unit test that will retrieve the metadata from the file.
Attempt to change the field using Jave on the Metadata class instance returned from box.

[sdk-docs]: ./doc

mattwiller:
Fixes #592


boxcla:

santoshaherkar:
When are we planning to make this fix available?

mattwiller:
@santoshaherkar This should go out with our scheduled release next week!

santoshaherkar:
I don't see support for test operation for multiselect fields. Was that kept out intentionally?

mattwiller:
@santoshaherkar The SDK should support testing an array element — can you share more details about what you're trying to do that isn't working?

santoshaherkar:
Currently there is no Metadata.test(path, List<String> values) method available to check if multiselect field had given values set.

PR #612, opened 2018-05-24
Fix check for duplicate As-User headers

### LINKED ISSUE #611: addHeader methot works wrong
https://github.com/box/box-java-sdk/blob/b881194f9a65cf44f9f400a72f7d9892e24e66dc/src/main/java/com/box/sdk/BoxAPIRequest.java#L168

You can't call indexOf(String) for List of RequestHeader, so this line always will return "-1" an user can add several "As-User" headers that will cause 403 response from server

mattwiller:
Fixes #611


boxcla:

PR #623, opened 2018-07-02
Sharedlink password

carycheng:


boxcla:

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -1394,6 +1394,52 @@ public void testUploadNewVersionReturnsCorrectInfo() throws IOException {
...
+                .willReturn(WireMock.aResponse()
+                        .withHeader("Content-Type", "application/json")
+                        .withStatus(200)));
+
+        WIRE_MOCK_CLASS_RULE.stubFor(WireMock.put(WireMock.urlPathEqualTo("/files/" + fileID))
+                .willReturn(WireMock.aResponse()
+                        .withHeader("Content-Type", "application/json")
+                        .withBody(result)));
+
+        WIRE_MOCK_CLASS_RULE.stubFor(WireMock.put(WireMock.urlPathEqualTo("/files/" + fileID))
```
mattwiller: Does this test actually make three API calls to the same URL?  Can you add comments to describe which calls these correspond to?

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -1394,6 +1394,52 @@ public void testUploadNewVersionReturnsCorrectInfo() throws IOException {
...
+                .willReturn(WireMock.aResponse()
+                        .withHeader("Content-Type", "application/json")
+                        .withBody(result)));
+
+        BoxFile file = new BoxFile(this.api, fileID);
+        BoxSharedLink.Permissions permissions = new BoxSharedLink.Permissions();
+
+        permissions.setCanDownload(true);
+        permissions.setCanPreview(true);
+        BoxSharedLink sharedLink = file.createSharedLink(BoxSharedLink.Access.OPEN, null, permissions);
```
mattwiller: Is it possible to allow setting the password in this call instead?  It would be much nicer for users to not have to make two API calls.

mattwiller:
Looks good — could you update the test to use the single-call way of setting the password to ensure that works correctly?

PR #662, opened 2018-11-15
Expose object type on BoxItem

### LINKED ISSUE #634: Type missing in BoxItem#Info
- [X ] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
BoxItem does not include `type` as an exposed field. This seems to be a major oversight - there is no way to check if an item is a file or folder. 

### Versions Used
Java SDK: 2.8.2
Java: 1.8.144

### Steps to Reproduce
Retrieve a BoxItem via an API call. There is no way to see what the item is except `getResource()` and performing a bunch of `isInstanceOf` checks... yikes.

[sdk-docs]: ./doc

mattwiller:
Fixes #634


boxcla:

PR #684, opened 2019-02-19
Added request id to exception message with trace id

carycheng:


boxcla:

src/main/java/com/box/sdk/BoxAPIResponseException.java
```diff
@@ -40,8 +40,16 @@ public BoxAPIResponseException(String message, BoxAPIResponse responseObj) {
         try {
             responseJSON = JsonObject.readFrom(responseObj.bodyToString());
 
-            if (responseObj.bodyToString() != null && responseJSON.get("request_id") != null) {
-                requestId = " | " + responseJSON.get("request_id").asString();
+            if (responseObj.bodyToString() != null && responseJSON.get("request_id") != null
```
mattwiller: You probably don't need to repeat the conditions from inside this branch here in the outer conditional — can you flatten this into two top-level conditions to construct `requestID`?

src/main/java/com/box/sdk/BoxAPIResponseException.java
```diff
@@ -40,16 +40,12 @@ public BoxAPIResponseException(String message, BoxAPIResponse responseObj) {
...
-                    requestId += responseJSON.get("request_id").asString();
-                }
+            if (responseObj.bodyToString() != null && responseJSON.get("request_id") != null) {
+                requestId += responseJSON.get("request_id").asString();
+            }
 
-                if (this.getHeaders().containsKey("BOX-REQUEST-ID")) {
-                    requestId += "." + this.getHeaders().get("BOX-REQUEST-ID").get(0).toString();
-                }
+            if (responseObj.bodyToString() != null && this.getHeaders().containsKey("BOX-REQUEST-ID")) {
```
mattwiller: I think you shouldn't need to check that the body is non-null here, since you're just reading out of the headers.  Is there a reason to keep this check here?

src/main/java/com/box/sdk/BoxAPIResponseException.java
```diff
@@ -40,16 +40,12 @@ public BoxAPIResponseException(String message, BoxAPIResponse responseObj) {
         try {
             responseJSON = JsonObject.readFrom(responseObj.bodyToString());
 
-            if (responseObj.bodyToString() != null && responseJSON.get("request_id") != null
-                    || this.getHeaders().containsKey("BOX-REQUEST-ID")) {
-
-                if (responseJSON.get("request_id") != null) {
-                    requestId += responseJSON.get("request_id").asString();
-                }
+            if (responseObj.bodyToString() != null && responseJSON.get("request_id") != null) {
```
mattwiller: What happens if `responseObj.bodyToString()` is null?  Wouldn't the statement above have already thrown in that case?  It might be helpful to have a test case with an empty body and the `BOX-REQUEST-ID` header to exercise that code path

carycheng: What is the expected behavior with empty body? The statement above does throw and we exit into the catch where we present the user with the message

`The API returned an error code <ERROR CODE>`

Wondering if that is the message we expect when there is an overall error reading the error response body or should we append the request id from the header into that error message as well? So something like this:

`The API returned an error code <ERROR CODE> | <REQUEST ID>`

Thank you @mattwiller !

PR #685, opened 2019-03-01
Wrong casting for response in BoxAPIResponse for BoxMetadataCascadePo…

carycheng:
…licy

Unfortunately, I introduced a bug in BoxMetadataCascadePolicy.forceApply().

The issue being this API does not return a response body. However, I casted the return type, originally being BoxAPIResponse to BoxJSONResponse. This would then throw a casting exception.

### DIFF
--- a/src/main/java/com/box/sdk/BoxMetadataCascadePolicy.java
+++ b/src/main/java/com/box/sdk/BoxMetadataCascadePolicy.java
@@ -145,7 +145,7 @@ public class BoxMetadataCascadePolicy extends BoxResource {
         JsonObject requestJSON = new JsonObject()
                 .add("conflict_resolution", conflictResolution);
         request.setBody(requestJSON.toString());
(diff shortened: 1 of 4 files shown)


boxcla:

src/main/java/com/box/sdk/BoxMetadataCascadePolicy.java
```diff
@@ -145,7 +145,7 @@ public void forceApply(String conflictResolution) {
         JsonObject requestJSON = new JsonObject()
                 .add("conflict_resolution", conflictResolution);
         request.setBody(requestJSON.toString());
-        BoxJSONResponse response = (BoxJSONResponse) request.send();
+        BoxAPIResponse response = request.send();
```
mattwiller: If you're not using the result anyway, why have it at all?  You could just do `request.send()` and throw away the result.

PR #697, opened 2019-04-22
Support for Set Metadata

carycheng:


boxcla:

doc/files.md
```diff
@@ -644,6 +645,21 @@ file.createMetadata(new Metadata().add("/foo", "bar"));
 [create-metadata-2]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFile.html#createMetadata-java.lang.String-com.box.sdk.Metadata-
 [create-metadata-3]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFile.html#createMetadata-java.lang.String-java.lang.String-com.box.sdk.Metadata-
 
+Set Metadata
```
mattwiller: Similar to how we've documented other set methods that wrap POST/PUT calls, I think we should move the existing "Create Metadata" and "Update Metadata" docs under this section as additional information.  I expect most people will want to use the new set method anyways, and we should make sure that they get that information up front.

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -1038,6 +1038,33 @@ public Metadata createMetadata(String typeName, String scope, Metadata metadata)
         return new Metadata(JsonObject.readFrom(response.getJSON()));
     }
 
+    /**
+     * Attempts to create a metadata value on a file. If conflict then perform an update of metadata values on file.
```
mattwiller: This description could probably be cleaned up to be more relevant for developers; it describes at a technical level what the method does, but I'm not sure that's what users are likely to care about.  Perhaps something like "Sets the provided metadata on the file, overwriting any existing metadata keys already present" could convey what someone should expect to happen as the end result of calling the method?

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -1038,6 +1038,33 @@ public Metadata createMetadata(String typeName, String scope, Metadata metadata)
...
+        Metadata metadataValue = null;
+
+        try {
+            metadataValue = this.createMetadata(templateName, scope, metadata);
+        } catch (BoxAPIException e) {
+            if (e.getResponseCode() == 409) {
+                Metadata metadataToUpdate = new Metadata(scope, templateName);
+                for (JsonValue value : metadata.getOperations()) {
+                    metadataToUpdate.add(value.asObject().get("path").asString(),
+                            value.asObject().get("value").toString());
```
mattwiller: This line worries me — does this work with all the data types we currently use for metadata?  In your tests, could you verify that this works correctly with numbers and arrays of strings (for `multiSelect` metadata)?

PR #703, opened 2019-05-21
Added test for making sure part buffer is correct size

carycheng:


boxcla:

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -1710,6 +1710,55 @@ public void testSetMetadataReturnsCorrectly() throws IOException {
...
+        final String uploadURL = "/files/upload_sessions/D5E3F8ADA11A38F0A66AD0B64AACA658";
+        final String commitURL = "/files/upload_sessions/D5E3F8ADA11A38F0A66AD0B64AACA658/commit";
+        FakeStream stream = new FakeStream("fakestream");
+
+        sessionResult = TestConfig.getFixture("BoxFile/CreateUploadSession201");
+        uploadResult1 = TestConfig.getFixture("BoxFile/UploadPartOne200");
+        uploadResult2 = TestConfig.getFixture("BoxFile/UploadPartTwo200");
+        commitResult = TestConfig.getFixture("BoxFile/CommitUpload201");
+
+        WIRE_MOCK_CLASS_RULE.stubFor(WireMock.post(WireMock.urlPathEqualTo(sessionURL))
```
mattwiller: Since I think this is currently the only unit test for Chunked Upload, would you mind adding constraints on the stubs for all four of these requests to make sure they're exactly what we expect them to be?  That way we get the maximum value from this test and could potentially catch other issues with Chunked Upload as well

mattwiller:
@carycheng Looks good to me!  Looks like you have an unused import to clean up, but after that feel free to merge 👍

PR #714, opened 2019-07-15
Added support for is_externally_owned field

carycheng:


boxcla:

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -1520,6 +1520,7 @@ public void removeWatermark() {
         private URL previewLink;
         private BoxLock lock;
         private boolean isWatermarked;
+        private boolean isExternallyOwned;
```
mattwiller: Should this be `Boolean` instead so it can represent a `null` value?

PR #721, opened 2019-08-01
Better exception handling for deserialization objects.

carycheng:


boxcla:

src/main/java/com/box/sdk/BoxDeserializationException.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Initializes the BoxDeserializationException class.
+     *
+     * @param member the key of the json member the deserialization occurred on.
+     * @param value the value of the json member the deserialization occurred on.
+     * @param className the name of the class the deserialization occurred on.
+     * @param e the throwable cause for the exception.
+     */
+    public BoxDeserializationException(String member, String value, String className, Exception e) {
```
mattwiller: Now that I think about it, the class where the exception was thrown from will be right at the top of the stack trace; we might want to remove the class name from the exception message.  That should clean up all the call sites where these exceptions are being constructed

src/main/java/com/box/sdk/BoxDeserializationException.java
```diff
@@ -0,0 +1,50 @@
...
+        super(e);
```
mattwiller: There's an `Exception` [constructor overload][ex-con] that takes both the message and the cause; it's probably better to use that than the `errorMessage` property, since that's not a standard exception field

[ex-con]: https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html#Exception(java.lang.String,%20java.lang.Throwable)

src/main/java/com/box/sdk/BoxWebHook.java
```diff
@@ -528,6 +528,8 @@ void parseJSONMember(JsonObject.Member member) {
                 assert false : "A ParseException indicates a bug in the SDK.";
             } catch (MalformedURLException e) {
                 assert false : "A MalformedURLException indicates a bug in the SDK.";
+            } catch (Exception e) {
```
mattwiller: You should remove the `catch` blocks above this so all exceptions get transformed into a `BoxDeserializationException`

PR #730, opened 2019-08-29
Set file attributes on chunked uploads

### LINKED ISSUE #729: Chunked upload does not allow setting file attributes
- [X] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
The uploadLargeFile() method for BoxFolder and BoxFile does not support setting file attributes, such as content_modified_at. The method should be updated so all supported file attributes can be set on the file during upload.

### Versions Used
Java SDK: 2.37.0
Java: 1.8

### Steps to Reproduce
Methods that support uploading large files do not accept file attributes

### Error Message, Including Stack Trace
N/A

[sdk-docs]: ./doc

stevebingsoo:
Added ability to set file attributes when uploading large files

Closes #729

### DIFF
--- a/doc/files.md
+++ b/doc/files.md
@@ -14,7 +14,10 @@ file's contents, upload new versions, and perform other common file operations
 - [Download a File](#download-a-file)
 - [Upload a File](#upload-a-file)
 - [Upload Preflight Check](#upload-preflight-check)
+- [Upload a Large File in Chunks](#upload-a-large-file-in-chunks)
+- [Upload a Large File in Chunks Including Attributes](#upload-a-large-file-in-chunks-including-attributes)
 - [Upload a Large File Version in Chunks](#upload-a-large-file-version-in-chunks)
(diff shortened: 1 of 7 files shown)


boxcla:
Verified that @stevebingsoo has signed the CLA. Thanks for the pull request!

mattwiller:
@stevebingsoo Thanks so much for the PR!  The code looks good, but we'll need to add test coverage to get the PR merged.  Unfortunately, the `IntegrationTest` tests don't count toward code coverage in CI; we can only run those tests locally because they require a specific live Box account.  If you can contribute a bit more of your time, you'll need to convert those tests all to `UnitTest`, which uses Wiremock to mock out the API and make the tests deterministic — examples throughout the test files.  Otherwise, our team on-call can pick the PR up from here and finish it up.  Which would you prefer?

carycheng:
Thank you @stevebingsoo for submitting this pull request and communicating with us, it is greatly appreciated!

We will do a release of the Java SDK this Thursday at which point this feature should be live, thanks again!

PR #733, opened 2019-09-11
Update files.md to tag thumbnail code sample

cbetta:


carycheng:
Hey @cbetta, looks good, curious about the format - should it be thumbnail or thumbnail_id? Wondering what the id refers to in this case, is it's just arbitrary then no worries, thanks!

cbetta:
Hey @carycheng, yes `get_files_id_thumbnail_id` is correct. 

In general terms, we replace every variable in the URL with a generic `id` and replace every forward slash and dot into a `_`.

In this case, that means `GET /files/:file_id/thumbnail.:extension` becomes `get_files_id_thumbnail_id`. The last ID is the extension.

More on this here: https://confluence.inside-box.net/display/PLT/Sample+Tags

PR #750, opened 2019-10-22
Update events.md

cbetta:
Update tags to mark both user and enterprise events


boxcla:
Verified that @cbetta has signed the CLA. Thanks for the pull request!

PR #760, opened 2019-12-13
SDK-1168 Metadata Query

PJSimon:


boxcla:
Verified that @PJSimon has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/MetadataTemplate.java
```diff
@@ -309,6 +314,55 @@ public static void deleteMetadataTemplate(BoxAPIConnection api, String scope, St
...
+                                            JsonArray orderBy, int limit, String marker) {
+
+        JsonObject jsonObject = new JsonObject().add("from", from);
+        if (query != null) {jsonObject.add("query", query);}
+        if (queryParameters != null) {jsonObject.add("query_params", queryParameters);}
+        if (ancestorFolderId != null) {jsonObject.add("ancestor_folder_id", ancestorFolderId);}
+        if (indexName != null) {jsonObject.add("use_index", indexName);}
+        if (orderBy != null) {jsonObject.add("order_by", orderBy);}
+        jsonObject.add("limit", limit);
+        if (marker != null) {jsonObject.add("marker", marker);}
```
sujaygarlanka: I don't think you need this line now.
```suggestion
```

src/main/java/com/box/sdk/BoxResourceIterable.java
```diff
@@ -68,28 +73,23 @@
...
-        this.api = api;
-        this.url = url;
-        this.limit = limit;
-        this.iterator = new IteratorImpl(null);
+        this(api, url, limit, null, null);
     }
 
     /**
-     * Constructor.
+     * Construtor.
```
sujaygarlanka: Constructor

src/main/java/com/box/sdk/MetadataTemplate.java
PJSimon: Yeah, the `metadata_queries/execute_read` doesn't take ANY query parameters.  Any passed in are ignored.  In order for for the marker to be respected, it has to be sent in the body.

src/main/java/com/box/sdk/BoxResourceIterable.java
PJSimon: How on Earth did you notice this?!?!  
Done!

src/main/java/com/box/sdk/MetadataTemplate.java
sujaygarlanka: Then the iterator has to be updated to send the marker in the body rather than the query as it is doing now.

PJSimon: It's doing both.  jsonObject is the body.  It gets passed in there and then again in the marker parameter.
I chose not to hard-code a null for the marker parameter so that any future changes will be supported.  (It's currently ignored).

sujaygarlanka: For the first request the marker is passed. For subsequent requests using the iterator, the next marker won't be passed in the body.

PJSimon: Good call!  Done!

src/main/java/com/box/sdk/BoxResourceIterable.java
```diff
@@ -163,7 +184,11 @@ private void loadNextPage() {
             QueryStringBuilder builder = new QueryStringBuilder(existingQuery);
             builder.appendParam(PARAMETER_LIMIT, BoxResourceIterable.this.limit);
             if (this.markerNext != null) {
-                builder.appendParam(PARAMETER_MARKER, this.markerNext);
+                if (this.body != null) {
+                    this.body.set("marker", this.markerNext);
```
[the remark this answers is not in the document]
PJSimon: Set appears to work.  This is tested when the third item is retrieved with a limit of 2 in the test.

Screen Shot 2019-12-17 at 10 35 38 AM

Screen Shot 2019-12-17 at 10 36 03 AM

PR #764, opened 2020-01-16
Fix Authentication Request Retries

PJSimon:


boxcla:
Verified that @PJSimon has signed the CLA. Thanks for the pull request!

src/main/java/com/box/sdk/BoxAPIRequest.java
```diff
@@ -666,9 +710,15 @@ void shouldAuthenticate(boolean shouldAuthenticate) {
         this.shouldAuthenticate = shouldAuthenticate;
     }
 
-    private static boolean isResponseRetryable(int responseCode) {
+    /**
+     *
+     * @param  responseCode HTTP error code of the response
+     * @return true if the response is one that should be retried, otherwise false
+     */
+    public static boolean isResponseRetryable(int responseCode) {
```
sujaygarlanka: May want to check for a 400 `invalid_grant` error where the server time differs from the local time. https://developer.box.com/en/guides/api-calls/permissions-and-errors/common-errors/

PJSimon: Yeah, per our discussion, I'm going to maintain parity with the existing SDK's functionality for now.  I created SDK-1262 for that.

PR #778, opened 2020-02-28
Return metadata for each item returned by a metadata query

sujaygarlanka:
`MetadataTemplate.executeMetadataQuery()` would returned `BoxItem` that did not include the metadata for each item. Now, instead of returning a `BoxItem`, the method returns a `BoxMetadataQueryItem`. This will contain an item and its associated metadata.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Next Release
+- Return metadata for each item returned by a metadata query
+
 ## 2.44.1 [2020-02-13]
 - Fix formatting bug for Java Logger
 - Improve date / time parsing for responses
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Next Release
-- Return metadata for each item returned by a metadata query
+- Add metadata to each item returned by a metadata query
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Next Release
+- Return metadata for each item returned by a metadata query
```
PJSimon:
```suggestion
- Add metadata to each item returned by a metadata query
```

src/main/java/com/box/sdk/BoxMetadataQueryItem.java
```diff
@@ -0,0 +1,93 @@
...
+
+import com.eclipsesource.json.JsonObject;
+import com.eclipsesource.json.JsonValue;
+
+/**
+ * Represents an individual item returned by a metadata query item.
+ *
+ * <p>Unless otherwise noted, the methods in this class can throw an unchecked {@link BoxAPIException} (unchecked
+ * meaning that the compiler won't force you to handle it) if an error occurs. If you wish to implement custom error
+ * handling for errors related to the Box REST API, you should capture this exception explicitly.*
```
PJSimon:
```suggestion
 * handling for errors related to the Box REST API, you should capture this exception explicitly.
```

Issue #777, opened 2020-02-18
BoxFile.rename method throws exception if debug logging is enabled


karolmarusiak:
- [x] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
An exception is throw when BoxFile.rename() method is called while logging level is set to FINE.

It looks like the the response stream is closed by the logging code and then the explicit call to disconnect (inside rename() method) attempts to read from this closed stream.

### Versions Used
Java SDK: 2.32.0
Java: 1.8

### Steps to Reproduce
Set the log level to FINE to print the response content and then call BoxFile.rename()

### Error Message, Including Stack Trace
Caused by: com.box.sdk.BoxAPIException: Couldn't finish closing the connection to the Box API due to a network error or because the stream was already closed.
	at com.box.sdk.BoxAPIResponse.disconnect(BoxAPIResponse.java:199)
	at com.box.sdk.BoxFile.rename(BoxFile.java:445)
	at com.appsense.box.impl.BoxDriveAPI$Transaction.<init>(BoxDriveAPI.java:331)
	... 27 more
Caused by: java.io.IOException: stream is closed
	at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.ensureOpen(HttpURLConnection.java:3417)
	at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.read(HttpURLConnection.java:3442)
	at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.read(HttpURLConnection.java:3437)
	at com.box.sdk.BoxAPIResponse.disconnect(BoxAPIResponse.java:189)
	... 29 more
[sdk-docs]: ./doc
[api-docs]: https://developer.box.com/docs

PJSimon:
@karolmarusiak, thanks for reporting this!  We'll take a look ASAP.

PR #779, opened 2020-02-28
Fix bug with premature disconnect in File Rename

PJSimon:
Resolves https://github.com/box/box-java-sdk/issues/777

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Next version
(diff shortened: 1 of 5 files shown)


sujaygarlanka:
Update changelog. Also does the folder item not need this fix?

PJSimon:
> Update changelog. Also does the folder item not need this fix?

No, the same issue in the folder class was actually fixed three years ago, this month!

PR #781, opened 2020-03-11
Check that part was successfully uploaded for large file uploads before retrying for 500 errors

sujaygarlanka:
The chunked upload endpoint that is hit by the `uploadLargeFile()` methods can sometimes return a 500 even though a part has been successfully uploaded. Previously, the SDK would simply re-upload the part after a 500 error. With this fix, the SDK will first check that the part was not successfully uploaded before retrying. This will reduce latency and network traffic.


src/main/java/com/box/sdk/BoxAPIRequest.java
```diff
@@ -404,6 +405,66 @@ public BoxAPIResponse send(ProgressListener listener) {
...
+      * <p>If the server returns an error code or if a network error occurs, then the request will be automatically
+      * retried. If the maximum number of retries is reached and an error still occurs, then a {@link BoxAPIException}
+      * will be thrown.</p>
+      *
+      * @param session The BoxFileUploadSession uploading the part
+      * @param offset Offset of the part being uploaded
+      * @throws BoxAPIException if the server returns an error code or if a network error occurs.
+      * @return A {@link BoxFileUploadSessionPart} part that has been uploaded.
+      */
+    public BoxFileUploadSessionPart sendForUploadPart(BoxFileUploadSession session, long offset) {
```
carycheng: If we don't want to expose this method and it's only meant for internal package use only for chunked upload we should consider different modifier, maybe default or protected?

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -1891,6 +1891,163 @@ public void testSetMetadataReturnsCorrectly() throws IOException {
...
+                .add(partOne);
+
+        JsonObject commitObject = new JsonObject()
+                .add("parts", parts);
+
+        WIRE_MOCK_CLASS_RULE.stubFor(WireMock.post(WireMock.urlPathEqualTo(sessionURL))
+                .withRequestBody(WireMock.equalToJson(sessionObject.toString()))
+                .willReturn(WireMock.aResponse()
+                        .withHeader("Content-Type", "application/json")
+                        .withBody(sessionResult)));
```
carycheng: What's the status code for this post to be returned?

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -1891,6 +1891,163 @@ public void testSetMetadataReturnsCorrectly() throws IOException {
...
+                        .withHeader("Digest", WireMock.containing("sha=31HjfCaaqU04+T5Te/biAgshQGw="))
+                        .withHeader("Content-Type", WireMock.containing("application/octet-stream"))
+                        .withHeader("Content-Range", WireMock.containing("bytes 0-4/5"))
+                        .withRequestBody(WireMock.equalTo("aaaaa"))
+                        .willReturn(WireMock.aResponse()
+                                .withHeader("Content-Type", "application/json")
+                                .withBody(responseBody500)
+                                .withStatus(500)));
+
+        WIRE_MOCK_CLASS_RULE.stubFor(WireMock.get(WireMock.urlPathEqualTo(listPartsURL))
```
carycheng: Here too

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -1891,6 +1891,163 @@ public void testSetMetadataReturnsCorrectly() throws IOException {
...
+                        .willReturn(WireMock.aResponse()
+                            .withHeader("Content-Type", "application/json")
+                            .withBody(partsResult)));
+
+        WIRE_MOCK_CLASS_RULE.stubFor(WireMock.post(WireMock.urlPathEqualTo(commitURL))
```
carycheng: Here as well

src/test/java/com/box/sdk/BoxFileTest.java
```diff
@@ -1891,6 +1891,163 @@ public void testSetMetadataReturnsCorrectly() throws IOException {
...
+                .add("offset", 0)
+                .add("size", 5);
+
+        JsonArray parts = new JsonArray()
+                .add(partOne);
+
+        JsonObject commitObject = new JsonObject()
+                .add("parts", parts);
+
+        WIRE_MOCK_CLASS_RULE.stubFor(WireMock.post(WireMock.urlPathEqualTo(sessionURL))
```
carycheng: This test needs to add the status codes for the stubbed api response

PR #785, opened 2020-03-23
adds shared link downscoping and coverage

boxdave:
- Adds string sharedLink arg thats a Box shared link to BoxAPIConnection.getLowerScopedToken. Returns a downscoped token for the resource associated to the shared link if the args not null.
- Adds test coverage.


src/test/java/com/box/sdk/BoxAPIConnectionTest.java
```diff
@@ -569,48 +569,65 @@ public void getLowerScopedTokenRefreshesTheTokenIfNeededbyCallingGetAccessToken(
         List<String> scopes = new ArrayList<String>();
         scopes.add("DummyScope");
         String resource = "";
+        String sharedLink = null;
 
         when(api.getTokenURL()).thenReturn("https://api.box.com/oauth2/token");
-        when(api.getLowerScopedToken(scopes, resource)).thenCallRealMethod();
+        when(api.getLowerScopedToken(scopes, resource, sharedLink)).thenCallRealMethod();
```
carycheng: will probably have to break this out into two tests - one for the original method and one for the new overloaded method

src/test/java/com/box/sdk/BoxAPIConnectionTest.java
```diff
@@ -569,48 +569,65 @@ public void getLowerScopedTokenRefreshesTheTokenIfNeededbyCallingGetAccessToken(
...
         try {
-            api.getLowerScopedToken(scopes, resource);
+            api.getLowerScopedToken(scopes, resource, sharedLink);
```
carycheng: same here

src/test/java/com/box/sdk/BoxAPIConnectionTest.java
```diff
@@ -569,48 +569,65 @@ public void getLowerScopedTokenRefreshesTheTokenIfNeededbyCallingGetAccessToken(
...
-        verify(api).getAccessToken();
+        ScopedToken token = api.getLowerScopedToken(scopes, resource, sharedLink);
+        assertThat(token, notNullValue());
+        assertThat(token.getAccessToken(), notNullValue());
     }
 
     @Test
     @Category(IntegrationTest.class)
-    public void getLowerScopedTokenWorks() {
+    public void getLowerScopedTokenForResource() {
```
carycheng: I think we can also add a unit test here with the intent to force it down the new if branches in boxapiconnection class. We can just have it return a string but we can assert that request to downscope token has all the parameter in the json body

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -677,7 +677,7 @@ public void setRequestInterceptor(RequestInterceptor interceptor) {
      * @param resource the resource for which the new token has to be obtained
```
carycheng: our linter is pretty strict with docstrings so with the overloaded method if you can put a @param sharedLink <sharedLink description> it will make then make the linter pass

src/test/java/com/box/sdk/BoxAPIConnectionTest.java
```diff
@@ -582,17 +579,56 @@ public void getLowerScopedTokenRefreshesTheTokenIfNeededbyCallingGetAccessToken(
 
     @Test
     @Category(UnitTest.class)
-    public void getLowerScopedTokenWithNullResource() {
+    public void getLowerScopedTokenForAPIEndpointResource() {
```
PJSimon, line 582: Actually, you can combine the following methods into one method, if you parameterize the `resource` string:
- getLowerScopedTokenForAPIEndpointResource
- getLowerScopedTokenForSharedLinkResource
- getLowerScopedTokenForNoteSharedLinkResource

PJSimon:
@boxdave I think the comments from Cary can be resolved, right?

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -671,11 +672,52 @@ public void setRequestInterceptor(RequestInterceptor interceptor) {
         this.interceptor = interceptor;
     }
 
+    private String determineResourceLinkType(String resourceLink) {
+
+        String resourceType = null;
```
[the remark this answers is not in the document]
dtrodger: @sujaygarlanka This is the enum I mentioned im looking for feedback on.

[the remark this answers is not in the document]
PJSimon: Per discussion, you can put the enum in this class

PR #787, opened 2020-04-01
Fix retry logic

sujaygarlanka:
Fix retry logic
- Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
- Honor Retry-After header if present by waiting for the time specified in the header before retrying
- Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 - Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
 - Honor Retry-After header, if present, by waiting for the time specified in the header before retrying
 - The concept of setting / getting "Maximum API Requests" has been deprecated in favor of "Maximum API Retries" to more clearly show the number of times a request will be retried after an error response is received.
-
 - Add ability to set expiration date for a collaboration ([#788](https://github.com/box/box-java-sdk/pull/788))
 - Add path parameter sanitization ([#790](https://github.com/box/box-java-sdk/pull/790))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,16 @@
 # Changelog
 
 ## Next Release
-- Fix bug with premature disconnect when renaming files and weblinks
-- Add metadata to each item returned by a metadata query
-- Check that part was successfully uploaded for large file uploads before retrying for 500 errors ([#781](https://github.com/box/box-java-sdk/pull/781))
-- Add preflight check before chunked uploads ([#782](https://github.com/box/box-java-sdk/pull/782))
 - Fix retry logic ([#787](https://github.com/box/box-java-sdk/pull/787))
   - Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
   - Honor Retry-After header if present by waiting for the time specified in the header before retrying
   - Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
+  
+## 2.45.0 [2020-04-02]  
+- Add preflight check before chunked uploads ([#782](https://github.com/box/box-java-sdk/pull/782))
+- Check that part was successfully uploaded for large file uploads before retrying for 500 errors ([#781](https://github.com/box/box-java-sdk/pull/781))
+- Fix bug with premature disconnect when renaming files and weblinks ([#779](https://github.com/box/box-java-sdk/pull/779))
+- Add metadata to each item returned by a metadata query ([#778](https://github.com/box/box-java-sdk/pull/778))
 
 ## 2.44.1 [2020-02-13]
 - Fix formatting bug for Java Logger
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## Next Release
 - Fix retry logic ([#787](https://github.com/box/box-java-sdk/pull/787))
-  - Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
-  - Honor Retry-After header if present by waiting for the time specified in the header before retrying
-  - Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
+- Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
+- Honor Retry-After header, if present, by waiting for the time specified in the header before retrying
+- Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
   
 ## 2.45.0 [2020-04-02]  
 - Add preflight check before chunked uploads ([#782](https://github.com/box/box-java-sdk/pull/782))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Add metadata to each item returned by a metadata query
 - Check that part was successfully uploaded for large file uploads before retrying for 500 errors ([#781](https://github.com/box/box-java-sdk/pull/781))
 - Add preflight check before chunked uploads ([#782](https://github.com/box/box-java-sdk/pull/782))
-- Fix retry logic ([#784](https://github.com/box/box-java-sdk/pull/784))
+- Fix retry logic ([#787](https://github.com/box/box-java-sdk/pull/787))
   - Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
   - Honor Retry-After header if present by waiting for the time specified in the header before retrying
   - Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - The concept of setting / getting "Maximum API Requests" has been deprecated in favor of "Maximum API Retries" to more clearly show the number of times a request will be retried after an error response is received.
 
 - Add ability to set expiration date for a collaboration ([#788](https://github.com/box/box-java-sdk/pull/788))
+- Add path parameter sanitization ([#790](https://github.com/box/box-java-sdk/pull/790))
 
 ## 2.45.0 [2020-04-02]
 - Add preflight check before chunked uploads ([#782](https://github.com/box/box-java-sdk/pull/782))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - Add metadata to each item returned by a metadata query
 - Check that part was successfully uploaded for large file uploads before retrying for 500 errors ([#781](https://github.com/box/box-java-sdk/pull/781))
 - Add preflight check before chunked uploads ([#782](https://github.com/box/box-java-sdk/pull/782))
+- Fix retry logic ([#784](https://github.com/box/box-java-sdk/pull/784))
+  - Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
+  - Honor Retry-After header if present by waiting for the time specified in the header before retrying
+  - Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
 
 ## 2.44.1 [2020-02-13]
 - Fix formatting bug for Java Logger
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Fix retry logic ([#787](https://github.com/box/box-java-sdk/pull/787))
 - Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
 - Honor Retry-After header, if present, by waiting for the time specified in the header before retrying
-- Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
+- The concept of setting / getting "Maximum API Requests" has been deprecated in favor of "Maximum API Retries" to more clearly show the number of times a request will be retried after an error response is received.
+
 - Add ability to set expiration date for a collaboration ([#788](https://github.com/box/box-java-sdk/pull/788))
 
 ## 2.45.0 [2020-04-02]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,14 @@
 # Changelog
 
+
 ## Next Release
 - Fix retry logic ([#787](https://github.com/box/box-java-sdk/pull/787))
 - Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
 - Honor Retry-After header, if present, by waiting for the time specified in the header before retrying
 - Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
-  
-## 2.45.0 [2020-04-02]  
+- Add ability to set expiration date for a collaboration ([#788](https://github.com/box/box-java-sdk/pull/788))
+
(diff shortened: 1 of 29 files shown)


src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -22,9 +22,9 @@
  */
 public class BoxAPIConnection {
     /**
-     * The default maximum number of times an API request will be tried when an error occurs.
+     * The default maximum number of times an API request will be retried after an error occurs.
      */
-    public static final int DEFAULT_MAX_ATTEMPTS = 5;
+    public static final int DEFAULT_MAX_RETRIES = 5;
```
PJSimon, line 27: Setting it to 5 makes the max total attempts 6, which is in line with the box-python-sdk (https://github.com/box/box-python-sdk/blob/9cdebdde39e68e735069dd97e0a459e09382455d/boxsdk/session/session.py#L341)

src/main/java/com/box/sdk/BoxAPIRequest.java
```diff
@@ -371,9 +371,9 @@ public BoxAPIResponse send() {
      */
     public BoxAPIResponse send(ProgressListener listener) {
         if (this.api == null) {
-            this.backoffCounter.reset(BoxGlobalSettings.getMaxRequestAttempts() + 1);
+            this.backoffCounter.reset(BoxGlobalSettings.getMaxRetryAttempts() + 1);
         } else {
-            this.backoffCounter.reset(this.api.getMaxRequestAttempts() + 1);
+            this.backoffCounter.reset(this.api.getMaxRetryAttempts() + 1);
```
PJSimon, line 376: Adding + 1 is the way to go, other wise we'd have to modify BackoffCounter.decrement() which seems too far.

src/main/java/com/box/sdk/BoxAPIConnection.java
sujaygarlanka: I like the idea of renaming this, but I think changing this value DEFAULT_MAX_RETRIES may be a breaking change. We are changing public values like the the functions to set and get the value. So if we are okay with our next release being a major release, then I am fine with this.

src/main/java/com/box/sdk/BoxAPIRequest.java
```diff
@@ -756,10 +764,13 @@ void shouldAuthenticate(boolean shouldAuthenticate) {
...
      *
      * @param  responseCode HTTP error code of the response
+     * @param  apiException BoxAPIException thrown
      * @return true if the response is one that should be retried, otherwise false
      */
-    public static boolean isResponseRetryable(int responseCode) {
-        return (responseCode >= 500 || responseCode == 429);
+    public static boolean isResponseRetryable(int responseCode, BoxAPIException apiException) {
+        String message = apiException.getMessage();
+        Boolean isClockSkewError =  responseCode == 400 && message.contains("exp");
```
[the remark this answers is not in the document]
sujaygarlanka: Going to fix this once we decide to rollback the `Max Request Attempts` commit or not.

src/main/java/com/box/sdk/BoxAPIConnection.java
PJSimon: Agreed.  I would think the chances someone is using that variable outside the SDK are very low, but if we deprecate it, then we can prevent the breaking change.  See next commit...

sujaygarlanka: Makes sense. I think we also need to deprecate `getMaxRequestAttempts` and `setMaxRequestAttempts` in `BoxAPIConnection` and `BoxGlobalSettings`. If we do this, it should be fine. The only other possible issue is the `restore` method in BoxAPIConnection. It takes in an object and restores settings. If someone is storing an object before this change, if they restore with that object there will be an issue. The old object will have the key `maxRequestAttempts`, but our new restore will only look for `maxRetryAttempts`. I don't think this will be a big issue, but not sure.

PJSimon: OK, I'll create a separate PR for this refactoring, then.  Thanks!

src/main/java/com/box/sdk/BoxAPIRequest.java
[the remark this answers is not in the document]
PJSimon: OK, lemme know when it's ready for final review.

src/main/java/com/box/sdk/BoxAPIRequest.java
```diff
@@ -769,7 +769,15 @@ void shouldAuthenticate(boolean shouldAuthenticate) {
      */
     public static boolean isResponseRetryable(int responseCode, BoxAPIException apiException) {
         String message = apiException.getMessage();
-        Boolean isClockSkewError =  responseCode == 400 && message.contains("exp");
+        String errorCode = "";
+        try {
+            JsonObject responseBody = JsonObject.readFrom(apiException.getResponse());
+            errorCode = responseBody.get("code").toString();
+        } catch (Exception e) { }
```
PJSimon, line 776: I believe in Java, it's better to check for existence of "code" than to rely on the overhead of throwing an exception, if it doesn't exist.

PJSimon: Do you still need the `try … catch` now?

sujaygarlanka: What if the readFrom fails, such as when the body is null on a 500 error?

PJSimon: We should check for that case, as well, then, for the same reason as above.

sujaygarlanka: Have to put the try catch back again because of a test case that wasn't passing [here](https://github.com/box/box-java-sdk/blob/69b2d57239ff41abcab895f5e0d30247ec39361b/src/test/java/com/box/sdk/BoxAPIResponseExceptionTest.java#L152). Apparently, an html body can be returned. In this case, I don't know how to do a check that the body is not HTML other than the try catch.

src/main/java/com/box/sdk/BoxAPIConnection.java
PJSimon: Did the other method deprecations you mentioned.
Also added a test to ensure `restore` works with "old" `state` strings.

src/test/Fixtures/BoxAPIConnection/State.json
```diff
@@ -0,0 +1 @@
+{"accessToken":"iKqkOC4FWkd1kG1HV9P17s5N7mqJBC2N","refreshToken":"QALVa7gLVz8bYf4B5wDYoXfhmR7746c4","lastRefresh":0,"expires":0,"userAgent":"Box Java SDK v2.45.0 (Java 1.8.0_131)","tokenURL":"https://api.box.com/oauth2/token","baseURL":"https://api.box.com/2.0/","baseUploadURL":"https://upload.box.com/api/2.0/","autoRefresh":true,"maxRetryAttempts":1}
```
PJSimon, line 1: This un-prettied format is the result of the call to save the state, so I preserved it in this test fixture.

src/test/java/com/box/sdk/BoxAPIConnectionTest.java
```diff
@@ -841,15 +859,15 @@ public String getJSON() {
     @Category(UnitTest.class)
     public void shouldUseGlobalMaxRequests() {
```
PJSimon, line 860:
```suggestion
    public void shouldUseGlobalMaxRetries() {
```

src/main/java/com/box/sdk/SharedLinkAPIConnection.java
```diff
@@ -93,14 +93,50 @@ public boolean getAutoRefresh() {
...
+     * are received.
+     * @param attempts the maximum number of request attempts.
+     * @deprecated setMaxRetryAttempts is preferred because it more clearly sets the number
+     * of times a request should be retried after an error response is received.
+     */
+    @Deprecated
     @Override
     public void setMaxRequestAttempts(int attempts) {
-        this.wrappedConnection.setMaxRequestAttempts(attempts);
+        this.wrappedConnection.setMaxRetryAttempts(attempts);
```
sujaygarlanka, line 119:
```suggestion
        this.wrappedConnection.setMaxRequestAttempts(attempts);
```

src/main/java/com/box/sdk/SharedLinkAPIConnection.java
```diff
@@ -93,14 +93,50 @@ public boolean getAutoRefresh() {
...
+     * are received.
+     * @return the maximum number of request attempts.
+     * @deprecated getMaxRetryAttempts is preferred because it more clearly gets the number
+     * of times a request should be retried after an error response is received.
+     */
+    @Deprecated
     @Override
     public int getMaxRequestAttempts() {
-        return this.wrappedConnection.getMaxRequestAttempts();
+        return this.wrappedConnection.getMaxRetryAttempts();
```
sujaygarlanka, line 106:
```suggestion
        return this.wrappedConnection.getMaxRequestAttempts();
```

src/main/java/com/box/sdk/BoxGlobalSettings.java
```diff
@@ -44,18 +44,44 @@ public static void setReadTimeout(int readTimeout) {
...
+     * are received.
+     * @param attempts maximum number of request attempts
+     * @deprecated setMaxRetryAttempts is preferred because it more clearly sets the number
+     * of times a request should be retried after an error response is received.
      */
-    public static void setMaxRequestAttempts(int maxRequestAttempts) {
-        BoxGlobalSettings.maxRequestAttempts = maxRequestAttempts;
+    @Deprecated
+    public static void setMaxRequestAttempts(int attempts) {
+        BoxGlobalSettings.maxRetryAttempts = attempts;
```
sujaygarlanka, line 67:
```suggestion
        BoxGlobalSettings.maxRetryAttempts = attempts - 1;
```

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -419,19 +429,45 @@ public boolean getAutoRefresh() {
...
+     * Sets the total maximum number of times an API request will be tried when error responses
+     * are received.
      * @return the maximum number of request attempts.
+     * @deprecated getMaxRetryAttempts is preferred because it more clearly gets the number
+     * of times a request should be retried after an error response is received.
      */
+    @Deprecated
     public int getMaxRequestAttempts() {
-        return this.maxRequestAttempts;
+        return this.maxRetryAttempts + 1;
```
sujaygarlanka, line 440:
```suggestion
        return this.maxRetryAttempts;
```

src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -22,10 +22,20 @@
...
 public class BoxAPIConnection {
     /**
-     * The default maximum number of times an API request will be tried when an error occurs.
+     * The default total maximum number of times an API request will be tried when error responses
+     * are received.
+     * @deprecated DEFAULT_MAX_RETRIES is preferred because it more clearly sets the number
+     * of times a request should be retried after an error response is received.
      */
+    @Deprecated
     public static final int DEFAULT_MAX_ATTEMPTS = 5;
```
sujaygarlanka, line 31: Leaving this in doesn't do anything because we don't use this anywhere. By deprecating it, we allow them to set it, but it doesn't affect functionality. So in the end, it is a breaking change unless we find a way to read this value.

src/main/java/com/box/sdk/BoxAPIConnection.java
sujaygarlanka: The + 1 should be removed because when our retry logic calls this method, it should return the number of retries. It will add 1 to the number of retries when it executes.

PJSimon: This method is deprecated, so the SDK doesn't call it anymore.  It's just preserved for client apps that may have been calling it.

PJSimon: This method is deprecated, so the SDK doesn't call it anymore. It's just preserved for client apps that may have been calling it.

src/main/java/com/box/sdk/BoxGlobalSettings.java
```diff
@@ -44,18 +44,44 @@ public static void setReadTimeout(int readTimeout) {
...
+     * Returns the global total maximum number of times an API request will be tried when error responses
+     * are received.
      * @return max number of request attempts
+     * @deprecated getMaxRetryAttempts is preferred because it more clearly gets the number
+     * of times a request should be retried after an error response is received.
      */
+    @Deprecated
     public static int getMaxRequestAttempts() {
-        return maxRequestAttempts;
+        return maxRetryAttempts;
```
PJSimon, line 55:
```suggestion
        return maxRetryAttempts + 1;
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 
 ## Next Release
+- Fix retry logic ([#787](https://github.com/box/box-java-sdk/pull/787))
+- Retry for 400 `invalid_grant` error in authentication requests (Clock Skew)
+- Honor Retry-After header, if present, by waiting for the time specified in the header before retrying
+- Number of request retries after the initial request is now equal to the `DEFAULT_MAX_ATTEMPTS` config value
```
PJSimon, line 8:
```suggestion
- The concept of setting / getting "Maximum API Requests" has been deprecated in favor of "Maximum API Retries" to more clearly show the number of times a request will be retried after an error response is received.

```

src/main/java/com/box/sdk/BoxAPIConnection.java
PJSimon: Correct, it's no longer used in the SDK.  It's deprecated because it's public, so it's possible that there are client apps out there reading this value from the SDK, for use in their client code.  

I don't understand your last sentence.

src/main/java/com/box/sdk/SharedLinkAPIConnection.java
PJSimon, line 106:
```suggestion
        return this.wrappedConnection.getMaxRetryAttempts() + 1;
```

PJSimon: Good point, but adding plus one to `getMaxRetryAttempts` is better because `getMaxRequestAttempts` is deprecated, now.

src/main/java/com/box/sdk/SharedLinkAPIConnection.java
PJSimon, line 119:
```suggestion
        this.wrappedConnection.setMaxRetryAttempts(attempts - 1);
```

PJSimon: Good point, but subtracting one is better because getMaxRequestAttempts is deprecated, now.

src/main/java/com/box/sdk/BoxAPIConnection.java
sujaygarlanka: You're right. Forgot it's static and can't be set.

Issue #770, opened 2020-01-25
Update Collaboration


yogaredhat:
Hi Team, 

I am following API document to update collaboration on folder .
https://developer.box.com/en/reference/put-collaborations-id/

Under request body  I do see option as expires_at but its not workign as it seems . I tried  API using curl and its not working. I tried SDK and I do not see any option to set expires_at date.

expires_at :string / date-timein bodyoptional
example2019-08-29T23:59:00-07:00
Set the expiration date for the collaboration. At this date, the collaboration will be automatically removed from the item.

Does this options still availble ?. If not is there any way we can expires collaboration for user other than explicity removing it.

carycheng:
Hey @yogaredhat,

In order to set expiration for a collaboration, the `expires_at` field needs to be set at the time of collaboration creation. Unfortunately, you cannot retroactively add it. In order to work around this you can delete the collaboration and re-add it with the `expires_at` field included.

Thank you

yogaredhat:
@carycheng  Thanks for this updates. Hoewever I do see any options to set  expires_at while adding collaboration. Here is how we are adding collaboration on folder. I might be missing something here  But I do not see any options where we can set expires_at. 

BoxFolder folder = new BoxFile(api, "id");
folder.collaborate("gcurtis@box.com", BoxCollaboration.Role.EDITOR);

sujaygarlanka:
@yogaredhat We currently don't have the option to set it when you create a collaboration in the Java SDK. We are looking into adding it in our next sprint.

sujaygarlanka:
@yogaredhat We have a feature in progress to set the expiration date when you create or update a collaboration in the Java SDK. The reason your curl call probably wasn't working earlier is because you must update your enterprise settings under `Content and Sharing` to allow owners to set the expiration date. I have attached a screen shot of where they can be found.
<img width="798" alt="Screen Shot 2020-04-07 at 4 19 10 PM" src="https://user-images.githubusercontent.com/21188938/78715544-8effd500-78eb-11ea-876f-2ef6c5289026.png">

Lastly, there is one quirk regarding this endpoint. When updating a collaboration, you must pass in both the role and the expiration date if you want to update the just expiration date.

PR #788, opened 2020-04-07
Add ability to set expiration date for a collaboration

sujaygarlanka:
Added the ability to set the expiration date when creating or updating a collaboration. Addresses this issue #770.


src/main/java/com/box/sdk/BoxCollaboration.java
```diff
@@ -86,6 +101,9 @@ public BoxCollaboration(BoxAPIConnection api, String id) {
         if (canViewPath != null) {
             requestJSON.add("can_view_path", canViewPath.booleanValue());
         }
+        if (expiresAt != null) {
```
PJSimon, line 104: This means you can't call this method to clear the expiration, right?  Is there another way to preserve the previous method signature and still allow for clearing the expiration?

sujaygarlanka: That's a good point. Just tested out sending null with Postman and the API doesn't accept it. Also tried an empty string and Null and no `expires_at` field. All options do not clear the `expires_at` field. So I think we may not have to have to allow the user to pass in null.
image

PR #790, opened 2020-04-08
Add path parameter sanitization

sujaygarlanka:
Added path parameter sanitization. So all IDs or other path parameters that should be numeric are checked to be numeric. If they are not, the SDK throws an error. In addition, all IDs that should be alphanumeric are checked to be alphanumeric. If they are not, the SDK throws an error.

### DIFF
--- a/src/main/java/com/box/sdk/BoxFile.java
+++ b/src/main/java/com/box/sdk/BoxFile.java
@@ -1051,7 +1051,7 @@ public class BoxFile extends BoxItem {
      * @return the metadata returned from the server.
      */
     public Metadata createMetadata(String typeName, String scope, Metadata metadata) {
-        URL url = METADATA_URL_TEMPLATE.build(this.getAPI().getBaseURL(), this.getID(), scope, typeName);
+        URL url = METADATA_URL_TEMPLATE.buildAlpha(this.getAPI().getBaseURL(), this.getID(), scope, typeName);
         BoxAPIRequest request = new BoxAPIRequest(this.getAPI(), url, "POST");
         request.addHeader("Content-Type", "application/json");
         request.setBody(metadata.toString());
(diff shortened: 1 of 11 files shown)


src/main/java/com/box/sdk/URLTemplate.java
```diff
@@ -18,12 +21,45 @@ public URLTemplate(String template) {
...
+                assert false : "An invalid path parameter passed in. It must be numeric.";
+            }
+        }
+        String urlString = String.format(base + this.template, values);
+
+        URL url = null;
+        try {
+            url = new URL(urlString);
+        } catch (MalformedURLException e) {
+            assert false : "An invalid URL template indicates a bug in the SDK.";
```
PJSimon, line 43:
```suggestion
            assert false : "A valid URL could not be constructed from the provided parameters.";
```

src/main/java/com/box/sdk/URLTemplate.java
```diff
@@ -37,13 +73,44 @@ public URL build(String base, Object... values) {
...
+            if (!NUMERIC.matcher(valueString).matches()) {
+                assert false : "An invalid path param passed in. It must be numeric.";
+            }
+        }
+        String urlString = String.format(base + this.template, values) + queryString;
+        URL url = null;
+        try {
+            url = new URL(urlString);
+        } catch (MalformedURLException e) {
+            assert false : "An invalid URL template indicates a bug in the SDK.";
```
PJSimon, line 94:
```suggestion
            assert false : "A valid URL could not be constructed from the provided parameters.";
```

src/main/java/com/box/sdk/URLTemplate.java
PJSimon: I don't think this is true, now that we have the new sanitization logic.  If there's a malformed URL, it's due to the parameters' values, right?

PJSimon: Also fix the other three instances of this verbiage in this class.

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -1051,7 +1051,7 @@ public Metadata createMetadata(String typeName, Metadata metadata) {
      * @return the metadata returned from the server.
      */
     public Metadata createMetadata(String typeName, String scope, Metadata metadata) {
-        URL url = METADATA_URL_TEMPLATE.build(this.getAPI().getBaseURL(), this.getID(), scope, typeName);
+        URL url = METADATA_URL_TEMPLATE.buildAlpha(this.getAPI().getBaseURL(), this.getID(), scope, typeName);
```
PJSimon, line 1054: If `this.getID()` returns, say, "12345abc", then this call would NOT return "An invalid path parameter passed in. It must be numeric." error message, even though that would be an invalid ID.

I can't think of any alternatives, but just calling out to start the discussion.

The API would handle the error in that case, so it's probably a minor concern, right?

sujaygarlanka: Ya, you are correct. There are ways around it like passing in what each parameter should be with each parameter in an Array for something like that. However, I think it's a minor concern and isn't worth making the SDK more complicated.

Issue #805, opened 2020-05-12
Calling isExternallyOwned() in BoxFolder.Info throws a NullPointerException


DaBearsCodeMonkey:
- [x] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [x] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported.
- [x] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
After creating a new folder, when I call getIsExternallyOwned(), I get a NullPointerException. All other methods work fine. 

### Versions Used
Java SDK: 1.8
Java: 1.8

### Steps to Reproduce

1. Instantiate a parent folder object
2. Create a new folder and accept the returned object as type BoxFolder.Info
3. Call getIsExternallyOwned()

### Error Message, Including Stack Trace
java.lang.NullPointerException: null
at com.box.sdk.BoxFolder$Info.getIsExternallyOwned(BoxFolder.java:1392) ~[box-java-sdk-2.41.0.jar:na]
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_232]
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_232]
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_232]
at java.lang.reflect.Method.invoke(Method.java:498) ~[na:1.8.0_232]
at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:688) ~[jackson-databind-2.9.8.jar:2.9.8]
at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:719) ~[jackson-databind-2.9.8.jar:2.9.8]
at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:155) ~[jackson-databind-2.9.8.jar:2.9.8]
at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serializeContents(IndexedListSerializer.java:119) ~[jackson-databind-2.9.8.jar:2.9.8]
[...]

PR #808, opened 2020-05-19
Fix issue for `getIsExternallyOwned()` for Files and Folders

sujaygarlanka:
Calling `getIsExternallyOwned()` on `File.Info` object or `Folder.info` would cause an exception when the value is null. This is because the return type for the method was a boolean, but the value being returned was a Boolean. So when the Boolean has value null, the value could not be cast to a primitive and would throw an exception. This issue was brought up in #805. 

This is fixed by changing the type of the value being returned from a Boolean to a boolean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Next Release
+- Fix issue for `getIsExternallyOwned()` for Files and Folders ([#808](https://github.com/box/box-java-sdk/pull/808))
+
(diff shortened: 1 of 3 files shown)


jacquesvandermerwe:
@sujaygarlanka, @PJSimon  whilst this works it now returns the wrong value. The results from the Java SDK vs. the CLI for the same folder return true (correct) in the CLI and 'false' (incorrect) in SDK 2.48.0. So no more nullpointer but the wrong result.

PR #809, opened 2020-05-19
Add support for the classification field for Files and Folders

sujaygarlanka:
Added classification field for `BoxFile.Info` and `BoxFolder.Info`.

### DIFF
--- a/src/main/java/com/box/sdk/BoxFile.java
+++ b/src/main/java/com/box/sdk/BoxFile.java
@@ -1768,8 +1768,8 @@ public class BoxFile extends BoxItem {
         }
 
         /**
-         * Gets the classification of this folder.
-         * @return the classification of this folder.
+         * Gets the metadata classification type of this file.
+         * @return the metadata classification type of this file.
          */
         public BoxClassification getClassification() {
             return this.classification;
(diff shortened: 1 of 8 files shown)


src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -1766,6 +1767,14 @@ public String getUploaderDisplayName() {
             return this.uploaderDisplayName;
         }
 
+        /**
+         * Gets the classification of this folder.
+         * @return the classification of this folder.
```
PJSimon, line 1772:
```suggestion
         * Gets the metadata classification type of this file.
         * @return the metadata classification type of this file.
```

src/main/java/com/box/sdk/BoxFolder.java
```diff
@@ -1398,6 +1399,14 @@ public boolean getIsExternallyOwned() {
             return this.isExternallyOwned;
         }
 
+        /**
+         * Gets the classification of this folder.
+         * @return the classification of this folder.
```
PJSimon, line 1404:
```suggestion
         * Gets the metadata classification type of this folder.
         * @return the metadata classification type of this folder.
```

PR #810, opened 2020-05-21
Add ability to set the filename when uploading a new version of a file

sujaygarlanka:
- Added the ability to set the filename when uploading a new version of a file.
- Fixed a bug with updating the `content_modified_at` field when upload a new version

### DIFF
--- a/src/main/java/com/box/sdk/BoxFile.java
+++ b/src/main/java/com/box/sdk/BoxFile.java
@@ -846,6 +846,20 @@ public class BoxFile extends BoxItem {
     }
 
     /**
+     * Uploads a new version of this file, replacing the current version. Note that only users with premium accounts
+     * will be able to view and recover previous versions of the file.
+     *
+     * @param fileContent     a stream containing the new file contents.
+     * @param fileContentSHA1 a string containing the SHA1 hash of the new file contents.
+     * @param modified        the date that the new version was modified.
+     * @param name            the new name for the file
+     * @return the uploaded file version.
+     */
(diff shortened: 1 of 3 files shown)


src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -875,6 +889,24 @@ public void uploadVersion(InputStream fileContent, String fileContentSHA1, Date
...
+     * @param fileContent     a stream containing the new file contents.
+     * @param fileContentSHA1 the SHA1 hash of the file contents. will be sent along in the Content-MD5 header
+     * @param modified        the date that the new version was modified.
+     * @param fileSize        the size of the file used for determining the progress of the upload.
+     * @param listener        a listener for monitoring the upload's progress.
+     * @param name            the new name for the file
+     * @return the uploaded file version.
+     */
+    public BoxFile.Info uploadNewVersion(InputStream fileContent, String fileContentSHA1, Date modified, long fileSize,
+                              ProgressListener listener, String name) {
```
PJSimon, line 909:
```suggestion
     * @param modified        the date that the new version was modified.
     * @param name            the new name for the file
     * @param fileSize        the size of the file used for determining the progress of the upload.
     * @param listener        a listener for monitoring the upload's progress.
     * @return the uploaded file version.
     */
    public BoxFile.Info uploadNewVersion(InputStream fileContent, String fileContentSHA1, Date modified, String name,
                              long fileSize, ProgressListener listener) {
```

PJSimon: I feel like the listener should always be last, no?

PR #811, opened 2020-05-27
Use double for metadata numbers

PJSimon:
Box JAVA SDK uses float to set value in metadata. Metadata uses number as field type. Metadata can handle values with higher precision as compared to JAVA float e.g. setting 233333333333333340 in number field returns 2333333341.

### DIFF
--- a/src/test/java/com/box/sdk/MetadataTest.java
+++ b/src/test/java/com/box/sdk/MetadataTest.java
@@ -196,6 +196,78 @@ public class MetadataTest {
 
     @Test
     @Category(IntegrationTest.class)
+    public void testMetadataPrecisionFloat() {
+        final float expectedValueFloat = 1234567890f;
+        BoxAPIConnection api = new BoxAPIConnection(TestConfig.getAccessToken());
+
+        long timestamp = Calendar.getInstance().getTimeInMillis();
+        String templateKey = "precision" + timestamp;
+        String fieldKey = "testPrecision";
+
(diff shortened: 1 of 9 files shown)


src/test/java/com/box/sdk/MetadataTest.java
```diff
@@ -194,6 +194,78 @@ public void testMetadataDate(String dateString, Long dateLong) {
...
+
+        // Add template to item
+        Metadata mdValues = new Metadata();
+        mdValues.add("/" + fieldKey, expectedValueFloat);
+        BoxFolder.Info folder = BoxFolder.getRootFolder(api).createFolder("Metadata Precision Test " + timestamp);
+        Metadata actualMD = folder.getResource().createMetadata(templateKey, mdValues);
+
+        Assert.assertEquals(templateKey, actualMD.getTemplateName());
+
+        final double actualValueDouble = actualMD.getFloat("/" + fieldKey);
```
sujaygarlanka, line 227: Should we deprecate the [getFloat()](https://github.com/box/box-java-sdk/blob/SDK-1404-use-double-for-metadata-numbers/src/main/java/com/box/sdk/Metadata.java#L347) and add a getDouble() method?

PJSimon:
Connection to Travis appears to have broken, but the last build passed: https://travis-ci.org/github/box/box-java-sdk/builds/695072611
Doing Admin Squash and Merge ...

PR #815, opened 2020-06-19
Clarify how to use orderBy parameter in executeMetadataQuery

randyjohnston:
Add orderBy and indexName parameters to `executeMetadataQuery(BoxAPIConnection api, String from, String query, JsonObject queryParameters, String ancestorFolderId, String indexName, JsonArray orderBy)` example to more closely mimic the REST API

### DIFF
--- a/doc/metadata_template.md
+++ b/doc/metadata_template.md
@@ -152,7 +152,7 @@ MetadataTemplate.deleteMetadataTemplate(api, "enterprise", "templateName");
 Execute Metadata Query
 --------------------------
 
-The `executeMetadataQuery(BoxAPIConnection api, String scope, String template)` queries files, folders and weblinks based on their metadata.
+The `executeMetadataQuery(BoxAPIConnection api, String from, String query, JsonObject queryParameters, String ancestorFolderId, String indexName, JsonArray orderBy)` method queries files and folders based on their metadata.
 
 <!-- sample post_metadata_queries_execute_read -->
 ```java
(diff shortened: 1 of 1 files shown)


doc/metadata_template.md
```diff
@@ -152,16 +152,21 @@ MetadataTemplate.deleteMetadataTemplate(api, "enterprise", "templateName");
...
 String ancestorFolderId = "0";
 JsonObject queryParameters = new JsonObject().add("arg", "test");
+JsonArray orderBy = new JsonArray();
+JsonObject primaryOrderBy = new JsonObject().add("field_key", "primarySortKey").add("direction", "asc");
+JsonObject secondaryOrderBy = new JsonObject().add("field_key", "secondarySortKey").add("direction",
+    "asc");
+orderBy.add(primaryOrderBy).add(secondaryOrderBy);
 
-BoxResourceIterable<BoxMetadataQueryItem> results = MetadataTemplate.executeMetadataQuery(api, from, query, queryParameters, ancestorFolderId);
+BoxResourceIterable<BoxMetadataQueryItem> results = MetadataTemplate.executeMetadataQuery(api, from, query, queryParameters, ancestorFolderId, null, orderBy);
```
cbetta, line 169: @randyjohnston what is this null here for? Is this the index name? Would it make sense to expand this example to just show how to use everything?

randyjohnston: Yes, it's the index name or `use_index` param in the POST request body. We may want to clarify that this is optional depending on the number of metadata template instances.

PR #819, opened 2020-06-25
Fix bug with setting the unshared at date for a shared link

mmarmol:
unshareAt field was not working, when you set it and try to create a sharelink or update one. The reason is simple, was using default java Date toString to convert to box date what is not correct. It needs to use BoxDateFormat instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Next Release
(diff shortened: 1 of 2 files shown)


PJSimon:
Hi @mmarmol ,

Thanks for this contribution! We will review it ASAP!

@PJSimon

sujaygarlanka:
@mmarmol This PR looks good! Thank you for submitting this. We appreciate you contributing to the Box Java SDK!

Issue #829, opened 2020-07-21
Issue deserializing JSON object when using SpringFox with Jackson


logicsec:
- [x] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [x] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [x] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
I am currently working on building a middleware application between our ERP and Box. We wanted to use the SDK but are running into an issue. The current setup is using Swagger and SpringFox. When we attempt to return a JSON object, it appears Jackson is having an issue deserializing the object and giving us the following error:

```java
{
  "timestamp": "2020-07-21T18:53:48.574+0000",
  "status": 500,
  "error": "Internal Server Error",
  "message": "Could not write JSON: The API returned an error code [404 | .038da827f44a95db6b57892565ac89f9d]; nested exception is com.fasterxml.jackson.databind.JsonMappingException: The API returned an error code [404 | .038da827f44a95db6b57892565ac89f9d] (through reference chain: com.box.sdk.BoxCollaboration$Info[\"createdBy\"]->com.box.sdk.BoxUser$Info[\"resource\"]->com.box.sdk.BoxUser[\"avatar\"])",
[...]

Issue #302, opened 2016-10-06
Info classes (subclasses of BoxJSONObject) are not de-serializable


dorfire:
They don't have public default constructors, and are encapsulated within another class, and thus **cannot be deserialized** with `com.fasterxml.jackson` and its likes.

gcurtis:
Hi @dorfire, thanks for the feedback. I have some questions about this issue.

1. Do you know of a way to fix this without making any breaking changes? If not, we can still investigate moving the inner classes out (or making them static) for the next major release.
2. If we move the classes out, they still need a reference to their `BoxResource` type, as is returned by `BoxUser.getResource()`, for example. It sounds like we would need to provide and empty constructor and just return null for that getter. Is that correct?
3. Are there specific areas where the default JSON deserialization logic is inadequate? Perhaps we could improve that to compensate for not being able to use Jackson.

PR #833, opened 2020-08-12
Add ability to get raw JSON from a response

richiehowelll:


sujaygarlanka:
@richiehowelll Is there a reason you would like to store the JSON object?

richiehowelll:
> @richiehowelll Is there a reason you would like to store the JSON object?

Yes, for caching purposes. You can see other issues this touches on such as #829 and #302 . Currently the domain objects in the SDK are not serializable which makes caching tricky. If we modify BoxJSONObject to preserve the raw JSON instead of throwing it away, it makes caching much simpler. The alternative would be to construct the JSON manually which gets complicated quickly when specific fields need to be updated.

src/main/java/com/box/sdk/BoxJSONObject.java
```diff
@@ -192,7 +197,7 @@ void update(JsonObject jsonObject) {
 
             this.parseJSONMember(member);
         }
-
+        this.jsonObject = jsonObject;
```
PJSimon, line 200: Any benefit to moving this up to the first line of the method?  I'm thinking that we'd gain the ability to save the jsonObject even if parseJSONMember throws an exception.

src/main/java/com/box/sdk/BoxUser.java
```diff
@@ -719,6 +719,14 @@ public Info() {
...
         }
 
+        /**
+         * Constructs an Info object by parsing information from a JSON string.
+         * @param  json the JSON string to parse.
+         */
+        public Info(String json) {
+            super(json);
+        }
+
```
PJSimon, line 729: Can you add this to the other two classes that were missing this constructor, too?
BoxFileUploadSession.Info
BoxWebHook.Info

src/main/java/com/box/sdk/BoxJSONObject.java
sujaygarlanka: Whenever update is called manually, the object will be saved if we put it update. If we move it up, it will only be called when instantiated.

PJSimon, line 200:
```suggestion
```

src/main/java/com/box/sdk/BoxJSONObject.java
```diff
@@ -181,7 +186,7 @@ void removePendingChange(String key) {
     }
 
     /**
-     * Updates this BoxJSONObject using the information in a JSON object.
+     * Updates this BoxJSONObject using the information in a JSON object and preserves the JSON object.
      * @param jsonObject the JSON object containing updated information.
      */
     void update(JsonObject jsonObject) {
```
PJSimon, line 192:
```suggestion
    void update(JsonObject jsonObject) {
        this.jsonObject = jsonObject;
        
```

PJSimon: I think if we move line 200 to here, then it's slightly more beneficial if the parseJSONMember call fails due to parsing exception.  At least in that case, the raw JSON is preserved.

src/main/java/com/box/sdk/BoxJSONObject.java
PJSimon: No, I'm not suggesting to move it out of the method.  See my code suggestion...

sujaygarlanka:
Thank you @richiehowelll for the PR!

PR #852, opened 2020-11-02
Fix for cross-enterprise collaborator calls to updateMetadata

AkihiroAsami:
When external users updates a metadata instance, the scope needs to be "enterprise_xxxx".
"xxxx" is the EID of the metadata template. The current code overwrites "enterprise_xxxx" to "enterprise".
This fix allows to keep the correct scope.

### DIFF
--- a/src/main/java/com/box/sdk/BoxFile.java
+++ b/src/main/java/com/box/sdk/BoxFile.java
@@ -1352,6 +1352,8 @@ public class BoxFile extends BoxItem {
         String scope;
         if (metadata.getScope().equals(Metadata.GLOBAL_METADATA_SCOPE)) {
             scope = Metadata.GLOBAL_METADATA_SCOPE;
+        } else if (metadata.getScope().startsWith(Metadata.ENTERPRISE_METADATA_SCOPE)) {
+            scope = metadata.getScope();
         } else {
             scope = Metadata.ENTERPRISE_METADATA_SCOPE;
         }


AkihiroAsami:
<why this fix is needed>
On the current version of Java SDK, you will get the error "tuple not found 404" when you execute the following sample code by the external user.
Because the scope will be changed from "enterprise_xxxx" to "enterprise"
by the current Java SDK code updateMetadata().
There is simple bug in this method's if clause as the following.
please fix this.

<the sample code>BoxFile file = new BoxFile(api,”123456789”);
Metadata meta = file.getMetadata(“templateName”, "enterprise_xxxx”);
meta.replace(“/foo”, “bar”);
file.updateMetadata(meta);

<the current if clause>
As-Is:
[if]
   the current scope is "global"  ==> it will be "global"
[else]
  the current scope is "enterprise_xxxx" ==> it will be "enterprise"  
  the current scope is in the other options ==> it will be "enterprise"

To-Be:
[if]
   the current scope is "global"  ==> it will be "global"
[else if]
  the current scope is "enterprise_xxxx" ==> it will be "enterprise_xxxx"
[else]
  the current scope is in the other options ==> it will be "enterprise"

PJSimon:
@AkihiroAsami Thanks for the PR!  It looks good to me, but I reached out to some other teams for feedback, to be sure.  I'm waiting to hear back.  I'll approve this PR ASAP!

PJSimon:
@AkihiroAsami Also, it turns out that we have a Java release scheduled in the next week or two, so we'll be able to push this out with minimal delay.

PJSimon:
@AkihiroAsami 
First, thanks for finding and proposing a fix for this bug.  You've exposed a shortcoming in this method that hasn't been discovered otherwise.  Thank you!

I did some research into this.  It looks this bug occurs when a collaborator is attempting to update metadata when the file is owned by an enterprise OTHER THAN the collaborator's enterprise.  That is the reason for passing in a scope like `enterprise_12345`, right?

Can you add this to the same method in BoxFolder, since it can occur there, as well?

Thanks!

src/main/java/com/box/sdk/BoxFile.java
```diff
@@ -1352,6 +1352,8 @@ public Metadata updateMetadata(Metadata metadata) {
         String scope;
         if (metadata.getScope().equals(Metadata.GLOBAL_METADATA_SCOPE)) {
             scope = Metadata.GLOBAL_METADATA_SCOPE;
+        } else if (metadata.getScope().startsWith(Metadata.ENTERPRISE_METADATA_SCOPE)) {
```
PJSimon, line 1355:
```suggestion
        } else if (metadata.getScope().startsWith(Metadata.ENTERPRISE_METADATA_SCOPE + "_")) {
```

PJSimon: This will make it just a little more explicit as to what we're looking for here...

PJSimon: Can you bring this logic over to BoxFolder, as well, please?

PJSimon:
@AkihiroAsami 
I'm not sure I can think of an easy way to write a test for this new logic, but if you'd like to add a test, that would be appreciated.

AkihiroAsami:
@PJSimon 

Thank you! The answer is here.

Q1.
It looks this bug occurs when a collaborator is attempting to update metadata when the file is owned by an enterprise OTHER THAN the collaborator's enterprise. That is the reason for passing in a scope like enterprise_12345, right?

A1.
That's right.

Q2.
- } else if (metadata.getScope().startsWith(Metadata.ENTERPRISE_METADATA_SCOPE
+ } else if (metadata.getScope().startsWith(Metadata.ENTERPRISE_METADATA_SCOPE + "_")) {

A2
Yes, I agree your suggestion.

Q3.
Can you bring this logic over to BoxFolder, as well, please?

A3.
BoxFolder is NOT needed this fix.
BoxFolder's updateMetadata() code is the following.
When composing the URL in this code, "metadata.getScope()" is used.
This means "scope" is not overwritten. There is no bug.

    //BoxFolder's updateMetadata method.
    public Metadata updateMetadata(Metadata metadata) {
        URL url = METADATA_URL_TEMPLATE.buildAlpha(this.getAPI().getBaseURL(), this.getID(), metadata.getScope(),
                metadata.getTemplateName());
        BoxAPIRequest request = new BoxAPIRequest(this.getAPI(), url, "PUT");
        request.addHeader("Content-Type", "application/json-patch+json");
        request.setBody(metadata.getPatch());
        BoxJSONResponse response = (BoxJSONResponse) request.send();
        return new Metadata(JsonObject.readFrom(response.getJSON()));
    }

If you think BoxFile and BoxFolder need to be created based on the same logic,
I think the following code is the ideal. What do you think of it ?

    //BoxFile's updateMetadata method.
    public Metadata updateMetadata(Metadata metadata) {
        URL url = METADATA_URL_TEMPLATE.buildAlpha(this.getAPI().getBaseURL(), this.getID(),metadata.getScope(),
                metadata.getTemplateName());
        BoxAPIRequest request = new BoxAPIRequest(this.getAPI(), url, "PUT");
        request.addHeader("Content-Type", "application/json-patch+json");
        request.setBody(metadata.getPatch());
        BoxJSONResponse response = (BoxJSONResponse) request.send();
        return new Metadata(JsonObject.readFrom(response.getJSON()));
    }

Q4
I'm not sure I can think of an easy way to write a test for this new logic, 
but if you'd like to add a test, that would be appreciated.

A4
In this case, should I add the test code on BoxFileTest.java ?

PJSimon:
@AkihiroAsami 
> Q3.
> Can you bring this logic over to BoxFolder, as well, please?

OK, let's NOT make the change in BoxFolder.  Let's only change `updateMetadata` in BoxFile.
We have to keep the `else` clause because removing it would be a breaking change for anyone passing in a bad scope value to the `updateMetadata` method.
I don't really agree with the `else` clause; I think the SDK should just use what ever scope is passed in from the client.  So, rather than bringing this logic around to all the CRUD methods for metadata in both BoxFile and BoxFolder, let's just make the fix in `updateMetadata` in BoxFile, today.

PJSimon:
@AkihiroAsami 

> Q4
> I'm not sure I can think of an easy way to write a test for this new logic,
> but if you'd like to add a test, that would be appreciated.
> 
> A4
> In this case, should I add the test code on BoxFileTest.java ?

Yes!  Thanks!

AkihiroAsami:
@PJSimon 
Thank you!

OK. I understand we have to consider the bad scope value.
In this perspective, you're right.
Let's go with "else" clause!

If making the test-code is NOT must needed, can I skip it?
Because my client needs this fix urgently.

PJSimon (approved):

PR #858, opened 2020-11-19
Fix bug with search including shared links

sujaygarlanka:
When hitting the search endpoint and setting `include_recent_shared_links` to true, a different response is returned than when it's false. Now a new model and method has been added to account for this difference.

### DIFF
--- a/src/test/Fixtures/BoxSearch/GetSearchItemsIncludingSharedLinks200.json
+++ b/src/test/Fixtures/BoxSearch/GetSearchItemsIncludingSharedLinks200.json
@@ -1,6 +1,7 @@
 {
   "entries": [
     {
+      "type": "search_result",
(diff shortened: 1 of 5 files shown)


src/test/Fixtures/BoxSearch/GetSearchItemsIncludingSharedLinks200.json
```diff
@@ -0,0 +1,86 @@
+{
+  "entries": [
+    {
+      "accessible_via_shared_link": "https://www.box.com/s/vspke7y05sb214wjokpk",
```
PJSimon, line 4: Assuming we're waiting for the `type` element to be implemented before adding here, right?
I see it's already added to `parseJSONMember` in BoxSearchSharedLink.java.

PJSimon (approved):

swfree (approved):

Issue #756, opened 2019-11-25
Weblink with Empty URL throwing the error


samrajcse:
Hi Team,

When we are listing down the folder we have weblink with empty url (I know we can't able to create now but might be created earlier). When I am iterating the Box folder api, it is throwing the below error. Same error in very latest API 2.41.0 too.

{
            "type": "web_link",
            "id": "<ID>",
            "sequence_id": "4",
            "etag": "4",
            "name": "<NAME>",
            "url": ""
        }

Exception stack trace:
		
Exception in thread "main" com.box.sdk.BoxAPIException: Couldn't parse url for weblink
	at com.box.sdk.BoxWebLink$Info.parseJSONMember(BoxWebLink.java:279)
	at com.box.sdk.BoxJSONObject.update(BoxJSONObject.java:193)
	at com.box.sdk.BoxJSONObject.<init>(BoxJSONObject.java:52)
	at com.box.sdk.BoxJSONObject.<init>(BoxJSONObject.java:42)
	at com.box.sdk.BoxResource$Info.<init>(BoxResource.java:175)
	at com.box.sdk.BoxItem$Info.<init>(BoxItem.java:234)
	at com.box.sdk.BoxWebLink$Info.<init>(BoxWebLink.java:245)
	at com.box.sdk.BoxItemIterator.next(BoxItemIterator.java:45)
	at com.box.sdk.BoxItemIterator.next(BoxItemIterator.java:8)
	at 
Caused by: java.net.MalformedURLException: no protocol: 
	at java.net.URL.<init>(URL.java:593)
	at java.net.URL.<init>(URL.java:490)
	at java.net.URL.<init>(URL.java:439)
	at com.box.sdk.BoxWebLink$Info.parseJSONMember(BoxWebLink.java:277)
	... 10 more

PJSimon:
Hi, @samrajcse!  Thanks for submitting this issue.  This week is the Thanksgiving holiday, so we will take a look next week and get back to you.

sujaygarlanka:
@samrajcse When a weblink is created, a valid url must be specified (https://developer.box.com/en/reference/post-web-links/). So when you get a weblink it should have a valid url. So if the url returned is invalid, the SDK will and should throw an error. In your case, an empty string is returned and is an invalid URL. To avoid getting the URL back, you can pass in a fields parameters. This will specify which fields to only be returned. For example, `https://api.box.com/2.0
[...]

PR #881, opened 2021-03-19
Fix `BoxWeblink` deserialization

sujaygarlanka:
Solves issue #756.

### DIFF
--- a/src/main/java/com/box/sdk/BoxWebLink.java
+++ b/src/main/java/com/box/sdk/BoxWebLink.java
@@ -275,7 +275,6 @@ public class BoxWebLink extends BoxItem {
             try {
                 if (memberName.equals("url")) {
                     try {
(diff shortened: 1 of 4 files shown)


src/main/java/com/box/sdk/BoxWebLink.java
```diff
@@ -277,7 +277,7 @@ protected void parseJSONMember(JsonObject.Member member) {
                     try {
                         this.linkURL = new URL(value.asString());
                     } catch (MalformedURLException e) {
-                        throw new BoxAPIException("Couldn't parse url for weblink", e);
+                        this.linkURL = null;
```
swfree, line 280: Should we only do this if the value is an empty string? i.e. set linkURL to null if the url is missing, but still throw the exception if a url is present but malformed

swfree (approved):

PR #890, opened 2021-05-11
Update execute metadata query to match API response

swfree:
### Issue Link :link:

- Fixes bug where `BoxResourceIterable<BoxMetadataQueryItem> results = MetadataTemplate.executeMetadataQuery(...)` was returning `null` results

### Implementation Details :construction:

- Removed the `executeMetadataQuery` that returns data as `BoxMetadataQueryItem`. Decided to remove instead of deprecate since the method does not work and always returns `null`
- Users should only use the `executeMetadataQuery` method that returns data as `BoxItem.Info`

### Testing Details :mag:

- Updated unit tests


sujaygarlanka (approved):

sujaygarlanka:
@swfree Just some linting errors that need to be fixed for Travis

swfree:
Thanks @sujaygarlanka! Fixed the linting errors

PR #891, opened 2021-05-12
Deprecate get thumbnail API in favor of get representations content

swfree:
### Goals :soccer:

- Deprecate getThumbnail method in favor of getRepresentationContent

### Testing Details :mag:

- Added an integration test for getRepresentationContent & ran unit tests


doc/files.md
```diff
@@ -761,21 +761,6 @@ URL embedLink = file.getPreviewLink();
 
 [get-preview-link]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFile.html#getPreviewLink--
 
-Get Thumbnail
```
sujaygarlanka, line 764: Do you think we should leave this, but indicate that it is deprecated? Also, include how a snippet from your test of how to do it with the get representation api?

swfree: @sujaygarlanka updated! get rep content is already documented (that's were I got the code snippet from 😄): https://github.com/box/box-java-sdk/blob/main/doc/files.md#get-representation-content

sujaygarlanka (approved):

PR #906, opened 2021-07-30
Fix FileRequest url format

mwwoda:
From owners of FileRequest API
`This field is intended to not have host name. As there are multiple scenarios for customers (app vs. ent and subdomains), these can change at runtime. We decided to not store or return hostname prior to release.`

### DIFF
--- a/src/main/java/com/box/sdk/BoxFileRequest.java
+++ b/src/main/java/com/box/sdk/BoxFileRequest.java
@@ -331,7 +331,7 @@ public class BoxFileRequest extends BoxResource {
         /**
          * Gets the title of file request.
          *
-         * @return the title of file request.`
+         * @return the title of file request.
          */
         public String getTitle() {
             return this.title;
--- a/src/main/java/com/box/sdk/BoxFileRequest.java
+++ b/src/main/java/com/box/sdk/BoxFileRequest.java
@@ -1,5 +1,6 @@
 package com.box.sdk;
 
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Date;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/box/sdk/BoxFileRequest.java
```diff
@@ -359,10 +358,11 @@ public Date getUpdatedAt() {
 
         /**
          * Gets the URL can be shared with users to let them upload files to the associated folder.
+         * The URL contains only the path (e.g. "/f/123456789").
          *
          * @return the date at which this task is due.
          */
-        public URL getUrl() {
+        public String getUrl() {
```
PJSimon, line 365: This is a breaking change, so I'm wondering if there's another approach we could take, for example generating the full URL from the partial path that comes back from the service.  Is there anyway to the get the first part of the URL by calling the Box API?  Or could a config parameter in the SDK be used (similar to `private static final String DEFAULT_BASE_URL = "https://api.box.com/2.0/";` here: https://github.com/box/box-java-sdk/blob/5e3a96c903fffa198c97e981ce75765a69bd6cb6/src/main/java/com/box/sdk/BoxAPIConnection.java#L43).  Maybe it could be inferred from a property of the BoxFolder?

mwwoda: This is breaking change from the SDK perspective, but according to service owners this change was done prior to the release. So technically, we could probably slide it off as a bugfix (incorrect model mapping), but if we don't want to break existing customers we could try to take the approach you proposed and build a full URL there.

- I don't think we have a possibility to get the first part of the URL from the Box API.
- I don't see any property that could be inferred from BoxFolder.
- We could use a config parameter, but we should add a new one as we don't have one for `https://app.box.com` used here. We should also make it configurable. According to the workflow team - the base URL can differ (different subdomains for enterprises for different customers). Maybe we can expose two properties here? Existing one with full URL and new one, only with path in case client needs to build their url differently at runtime.

mwwoda: Decided to use the third approach with config parameter while still giving ability to change the base url manually.

mhagmajer (commented):
lgtm

Will leave for @PJSimon to sign off on this

PJSimon (approved):

PR #907, opened 2021-08-17
Replace insensitive language (submaster, MASTER_INVITE_ACCEPT and MASTER_INVITE_REJECT)

mwwoda:


src/main/java/com/box/sdk/BoxGroupMembership.java
```diff
@@ -352,6 +381,60 @@ String toJSONString() {
         }
     }
 
+    /**
+     * Enumerates the possible roles that a user can have within a group.
+     */
+    public enum MembershipRole {
```
PJSimon, line 387: Do you think GroupRole iis a (slightly) better name here?

src/main/java/com/box/sdk/BoxGroupMembership.java
```diff
@@ -158,20 +163,41 @@ public Info(String json) {
...
         public void setRole(Role role) {
             this.role = role;
             this.addPendingChange("role", role.toJSONString());
         }
 
+        /**
+         * Gets the level of access the user has.
+         * @return the level of access the user has.
+         */
+        public MembershipRole getMembershipRole() {
```
PJSimon, line 188:
```suggestion
        public MembershipRole getRole() {
```
This is less of a change, but does it really make things that much easier for a client adopting the change if we preserve the name of the deprecated method?

src/main/java/com/box/sdk/BoxGroupMembership.java
```diff
@@ -158,20 +163,41 @@ public Info(String json) {
...
+            return this.membershipRole;
+        }
+
+        /**
+         * Sets the level of access the user has.
+         * @param role the new level of access to give the user.
+         */
+        public void setMembershipRole(MembershipRole role) {
```
PJSimon, line 196:
```suggestion
        public void setRole(MembershipRole role) {
```
This is less of a change, but does it really make things that much easier for a client adopting the change if we preserve the name of the deprecated method?

src/test/java/com/box/sdk/BoxGroupMembershipTest.java
```diff
@@ -97,7 +100,8 @@ public void testUpdateInfoParseAllFieldsCorrectly() throws ParseException {
         api.setRequestInterceptor(JSONRequestInterceptor.respondWith(fakeJSONResponse));
         BoxGroupMembership membership = new BoxGroupMembership(api, id);
         BoxGroupMembership.Info info = membership.new Info();
-        info.setRole(Role.SUBMASTER);
+        Role responseRole = info.getRole();
```
PJSimon, line 103: responseRole not used

src/main/java/com/box/sdk/BoxGroupMembership.java
mwwoda: We can't preserve the getRole name in the new implementation. Java doesn't allow that even if they have a different return type. They would need to have different name and/or parameters.

src/main/java/com/box/sdk/BoxGroupMembership.java
mwwoda: Because of the comment for the getRole() I think we should also preserve this name to be consistent with get method name.

src/main/java/com/box/sdk/BoxGroupMembership.java
mwwoda: yep, definitely better

src/test/java/com/box/sdk/BoxGroupMembershipTest.java
mwwoda: Removed

src/main/java/com/box/sdk/BoxEvent.java
```diff
@@ -167,8 +177,16 @@ void parseJsonMember(JsonObject.Member member) {
                 }
             }
 
-            if (this.type == null) {
+            for (EventType t : EventType.values()) {
+                if (t.name().equals(stringValue)) {
+                    this.eventType = t;
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
mwwoda: I think using `valueOf` is not the best approach here. If we don't find the event in our `EventType` enum we return `EventType.UNKNOWN`. Controlling flow of the application in catch clauses is often considered as anti-pattern. Instead, I reimplemented `valueOf` that will return `null` instead of throwing an exception if value is not found. It's still O(1) like the original `valueOf`.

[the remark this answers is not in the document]
PJSimon: Yes, I was thinking that the exception would happen so extremely rarely, that it would be acceptable, but actually, that may not be true for old versions of the SDK that are still in use, with an out-of-date EventType enum.

PJSimon (approved):

PR #939, opened 2021-11-18
Documentation changes

antusus:


mwwoda (approved):

jacquesvandermerwe (commented):
Looking good.

In the code sample the section:
```
for (BoxEvent event : eventLog){
  System.out.println("Enterprise Event Created by User: "
    + event.getCreatedBy().getName()
    + " Login: " + event.getCreatedBy().getLogin()
    + " Event Type: " + event.getEventType()
    + " Created at: " + event.getCreatedAt().toString()
  );
};
```
the .getType is deprecated. There is a method called .getEventType.

General observation on the Live monitoring, questions we may get.  

- If I get the next stream position, and make a subsequent call, will the API block until an event is returned ie. Result or will it immediately return with no records and the next stream position? 

- Is there a max limit that cannot  be exceeded ie. 1000 so any limit greater than will be ignored?

- If I have a next stream position (which I created in the past) and it's older than 2 weeks, I assume that it will only return what it has access to?

- In the comment where it says on an Enterprise admin account can make the call, people may confuse it with a user, but a service account with manage enterprise properties can also get the event stream ? Correct?

antusus:
> * In the comment where it says on an Enterprise admin account can make the call, people may confuse it with a user, but a service account with manage enterprise properties can also get the event stream ? Correct?

Yes this is correct. I'll update docs.

PR #940, opened 2021-11-22
Changed BoxDeveloperEditionAPIConnection#getAppUserConnection to BoxDeveloperEditionAPIConnection#getUserConnection.

antusus:


jacquesvandermerwe (commented):
For this section in the authentication.md, unless that will be auto generated from the .java
`It is also possible to get an API connection for an app user by doing something like this:`

Could you add this:
`...API connection for an app user or managed user...`

arjankowski (approved):
👍

PR #947, opened 2021-11-30
feat: Adding `BoxFile#getVersions(String... fields)` to allow users to specify what information they want to extract. Fixes #946.

### LINKED ISSUE #946: File Versions: Unable to get "version_number" for versions
### Is your feature request related to a problem? Please describe.
A the moment we can get the versions of a file using the following snippet. 
```
BoxFile file = new BoxFile(api, "id");
Collection<BoxFileVersion> versions = file.getVersions();
for (BoxFileVersion version : versions) {
    System.out.format("SHA1 of \"%s\": %s\n", item.getName(), version.getSha1());
}
```

The issue is that the BoxFileVersion object doesn't have the version number of that particular version. In order to get it we need to send the fields parameter just like we do with getInfo(fields) but that is not available. At the moment to get that information I have to create a call like this:
```
            URL url = new URL(String.format("https://api.box.com/2.0/files/%s/versions/%s?fields=id,type,name,created_at,version_number,ext",objectID,version.getVersionID()));
            BoxAPIRequest request = new BoxAPIRequest(localApi, url, "GET");
            BoxJSONResponse response = (BoxJSONResponse) request.send();
            String json = response.getJSON();
```

As you can see 'version_number' is one of the fields requested

### Describe the solution you'd like
It would be good to have a fields option with the 
`file.getVersions()` ie. `file.getVersions(fields)` where we can then define the fields to be returned.

### Describe alternatives you've considered
The pure REST call as described above.

antusus:
We want to add possibility for customers to get specific fields when asking for file versions.
One of the fields requested is version_number. It is returned by API as a string. It will always contain a number.
Customers can se a limit to 1-100 or mark that they want to keep all versions of the file. Maximum allowed version is: 4,294,967,295.
I decided that we should return version number as a numeric value.  To cover the range I've picked Long which can hold up to 2^63 -1.
However if we feel like we should return a String that is also possible.


src/main/java/com/box/sdk/BoxFileVersion.java
```diff
@@ -252,6 +256,15 @@ public Date getPurgedAt() {
         return this.purgedAt;
     }
 
+    /**
+     * Returns version number.
+     *
+     * @return String representing version number
```
lukaszsocha2, line 262: Long not String

lukaszsocha2 (dismissed):
Just one comment fix should be done

lukaszsocha2 (approved):

PR #1004, opened 2022-02-15
chore: documentation update with CCG and extending retention on a file

antusus:


doc/authentication.md
```diff
@@ -156,8 +155,47 @@ BoxAPIConnection api = BoxTransactionalAPIConnection.getTransactionConnection("Y
...
 ```
 
-Manual Token Creation
----------------------
+### Client Credentials Grant
+Allows you to obtain an access token by having client credentials and secret with enterprise or user ID, 
+which allows you to work using service or user account.
+
+The `BoxCCGAPIConnection` works the same way as the `BoxAPIConnection` so for example to get root folder you can do:
+```java
```
mgrytsai, line 163: Wouldn't we need to include tag `<!-- sample x_auth with_client_credentials -->` to make sure sample code is shown in developer website?

mgrytsai (approved):

PR #1005, opened 2022-02-16
chore: Run integration tests on `CI`

antusus:


src/intTest/java/com/box/sdk/BoxRecentsIT.java
```diff
@@ -9,11 +11,13 @@
 import java.util.Iterator;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
 /**
  * {@link BoxRecents} related integration tests.
  */
+@Ignore
```
mwwoda, line 20: I suppose that there is a problem with this test, right?

src/intTest/java/com/box/sdk/EventStreamIT.java
```diff
@@ -6,8 +6,10 @@
 
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
+import org.junit.Ignore;
 import org.junit.Test;
 
+@Ignore
```
mwwoda, line 12: I suppose that there is a problem with this test, right?

src/main/java/com/box/sdk/BoxFileVersionRetention.java
```diff
@@ -67,6 +67,7 @@ public BoxFileVersionRetention(BoxAPIConnection api, String id) {
      * and {@link BoxRetentionPolicyAssignment#getFileVersionsUnderRetention(String...)} instead.
      * Retrieves all file version retentions matching given filters as an Iterable.
      */
+    @Deprecated
```
mwwoda, line 70: Is there any alternative to this method? Is it possible to point a client to it?

mwwoda (dismissed):
Approving with small comments/questions.

src/main/java/com/box/sdk/BoxFileVersionRetention.java
antusus: Yes, but it is documented in JavaDoc above the method:

```
* @deprecated This method will be deprecated in the future. Please use
* {@link BoxRetentionPolicyAssignment#getFilesUnderRetention(int, String...)}
* and {@link BoxRetentionPolicyAssignment#getFileVersionsUnderRetention(String...)} instead.
```

src/intTest/java/com/box/sdk/BoxRecentsIT.java
antusus: I forgot to enable it. But it required user connection.

src/intTest/java/com/box/sdk/EventStreamIT.java
antusus: I forgot to enable that.

mwwoda (dismissed):

mwwoda (dismissed):

PR #1014, opened 2022-02-18
fix: Documentation Client credentials grant sample code tag

mgrytsai:
Correct sample code tag, so it [developer.box.com](https://developer.box.com/guides/authentication/client-credentials/) correctly display sample code.


antusus (approved):
👍

PR #1017, opened 2022-02-21
chore: Cleanup unit tests

antusus:
- chore: removing WireMockClassRule
- chore: removing thread sleeps from unit tests

### DIFF
--- a/src/test/java/com/box/sdk/BoxAPIResponseExceptionTest.java
+++ b/src/test/java/com/box/sdk/BoxAPIResponseExceptionTest.java
@@ -4,17 +4,19 @@ import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig;
+import static java.lang.String.format;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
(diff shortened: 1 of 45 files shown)


src/test/java/com/box/sdk/BoxEventTest.java
```diff
@@ -1,26 +1,34 @@
...
 
-    @ClassRule
-    public static final WireMockClassRule WIRE_MOCK_CLASS_RULE = new WireMockClassRule(53621);
+    @Rule
+    public WireMockRule wireMockRule = new WireMockRule(wireMockConfig().dynamicPort());
     private final BoxAPIConnection api = TestConfig.getAPIConnection();
 
+    @Before
+    public void setUpBaseUrl() {
+        api.setBaseURL(baseUrl());
```
mwwoda, line 29: Every other `setUpBaseUrl()` method also includes `api.setMaxRetryAttempts(1)` and this one doesn't. I'm not sure if it was an oversight or was it intentional. Have you considered creating some sort of test base class or moving this logic to a separate function to ensure that the same logic is called every time?

src/test/java/com/box/sdk/BoxAPIResponseExceptionTest.java
```diff
@@ -4,41 +4,45 @@
...
-
     @Rule
-    public WireMockRule wireMockRule = new WireMockRule(53620);
+    public WireMockRule wireMockRule = new WireMockRule(wireMockConfig().dynamicPort());
 
     private final BoxAPIConnection api = TestConfig.getAPIConnection();
 
+    @Before
+    public void setUpBaseUrl() {
+        api.setBaseURL(baseUrl());
```
mwwoda, line 43: Every other `setUpBaseUrl()` method also includes `api.setMaxRetryAttempts(1)` and this one doesn't. I'm not sure if it was an oversight or was it intentional. Have you considered creating some sort of test base class or moving this logic to a separate function to ensure that the same logic is called every time?

mwwoda (dismissed):

src/test/java/com/box/sdk/BoxEventTest.java
antusus: Thanks. I'll set this.

PR #1023, opened 2022-02-25
chore: added publication to Box Artifactory

antusus:


antusus:
We are not going to publish to releases repository

PR #1056, opened 2022-06-02
docs: Update metadata documentation for a file and folder

lukaszsocha2:
Closes: SDK-2260

### DIFF
--- a/doc/files.md
+++ b/doc/files.md
@@ -821,7 +821,8 @@ Set Metadata
 
 To set metadata on a file, call [`setMetadata(String templateName, String scope, Metadata metadata)`][set-metadata].
 This method will try to create provided metadata on a file. However, if metadata has already been applied to this file,
-it will overwrite values of metadata keys specified in the `metadata` parameter.
+it will overwrite values of metadata keys specified in the `metadata` parameter. The metadata keys not specified in
+the `metadata` parameter will remain unchanged.
 
 ```java
 BoxFile file = new BoxFile(api, "id");
--- a/doc/files.md
+++ b/doc/files.md
@@ -8,7 +8,6 @@ file's contents, upload new versions, and perform other common file operations
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
 - [Get a File's Information](#get-a-files-information)
 - [Update a File's Information](#update-a-files-information)
 - [Download a File](#download-a-file)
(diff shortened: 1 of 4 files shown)


doc/files.md
```diff
@@ -820,30 +807,39 @@ BoxFile file = new BoxFile(api, "id");
...
+[create-metadata-2]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFile.html#createMetadata-java.lang.String-com.box.sdk.Metadata-
+[create-metadata-3]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFile.html#createMetadata-java.lang.String-java.lang.String-com.box.sdk.Metadata-
+[update-metadata]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFile.html#updateMetadata-com.box.sdk.Metadata-
+
+Set Metadata
+------------
+
+To set metadata on a file, call [`setMetadata(String templateName, String scope, Metadata metadata)`][set-metadata].
+This method will try to create provided metadata on a file. However, if metadata has already been applied to this file,
+it will overwrite values of metadata keys specified in the `metadata` parameter.
```
antusus, line 824: Should we say that it will execute 2 API calls?
Should we say even clearer that if you do not change metadata fields they will remain unchanged? It says it will change values specified but some users were confused.

doc/folders.md
```diff
@@ -425,49 +426,50 @@ Collection<BoxCollaboration.Info> collaborations = folder.getCollaborations();
...
+[create-metadata]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFolder.html#createMetadata-com.box.sdk.Metadata-
+[create-metadata-2]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFolder.html#createMetadata-java.lang.String-com.box.sdk.Metadata-
+[create-metadata-3]: http://opensource.box.com/box-java-sdk/javadoc/com/box/sdk/BoxFolder.html#createMetadata-java.lang.String-java.lang.String-com.box.sdk.Metadata-
 
-Note: This method will only succeed if the provided metadata template is not currently applied to the folder, otherwise
-it will fail with a Conflict error.
+Set Metadata
+------------
+
+To set metadata on a folder, call [`setMetadata(String templateName, String scope, Metadata metadata)`][set-metadata].
```
antusus, line 454: In files we have this part added:
```
This method will try to create provided metadata on a file. However, if metadata has already been applied 
to this file, it will overwrite values of metadata keys specified in the `metadata` parameter.
```
Maybe we could copy it here as well?
Should we say that it will execute 2 API calls?
Should we say even clearer that if you do not change metadata fields they will remain unchanged? It says it will change values specified but some users were confused.

doc/files.md
lukaszsocha2: `Should we say that it will execute 2 API calls?`
I think no because sometimes this method can also execute only 1 call.

lukaszsocha2: `Should we say even clearer that if you do not change metadata fields they will remain unchanged?`
I added that part

antusus (approved):
👍

PR #1072, opened 2022-07-06
feat: Added pagination for getting items in trash with new `BoxTrash#items` API

antusus:


src/main/java/com/box/sdk/BoxTrash.java
```diff
@@ -247,7 +247,38 @@ public BoxFile.Info restoreFile(String fileID, String newName, String newParentI
...
+     * @param pagingParameters describes paging parameters
+     * @param fields           the fields to retrieve.
+     * @return an iterable containing the items in the trash.
+     */
+    public Iterable<BoxItem.Info> items(
+        SortParameters sortParameters,
+        PagingParameters pagingParameters,
+        String... fields
+    ) {
+        QueryStringBuilder builder = sortParameters.asQueryStringBuilder();
```
arjankowski, line 269: I tested this locally with code snippet from documentation:
```
BoxTrash trash = new BoxTrash(api);
Iterable<BoxItem.Info> trashEntries = trash.items(
  SortParameters.ascending("name"),
  PagingParameters.marker(500)
);
```
and unfortunately I've received 400 status code with response:

```
{"type":"error","status":400,"code":"bad_request","context_info":{"errors":[{"reason":"invalid_parameter","name":"sort","message":"The sort parameter is not allowed in marker-based paginaiton for this endpoint"}]},"help_url":"http:\/\/developers.box.com\/docs\/#errors","message":"Bad Request","request_id":"..."}
```
Shouldn't we disallow to use sort parameter when using marker-based?
Actually documentation mentioned it here https://developer.box.com/reference/get-folders-trash-items/#param-sort -> `This parameter is not supported when using marker-based pagination.`

arjankowski (approved):
👍

PR #1073, opened 2022-07-06
chore: Documentation changes for the next release

antusus:


doc/trash.md
```diff
@@ -29,11 +29,28 @@ the collection of items currently in the trash, simply iterate over it.
 ```java
```
arjankowski, line 29: I've checked that java sample is not present on https://developer.box.com/reference/get-folders-trash-items/.
This is probably because of invalid sample tag. I think it should be `get_folders_trash_items` instead of `get_trash`

arjankowski (dismissed):
👍

arjankowski (commented):
to turn off

doc/folders.md
```diff
@@ -121,26 +121,25 @@ long offset = 0;
...
     // Do something
 }
 ```
 
 With offset pagination you cannot set offset larger than 300000.
-
-By default, SDK is using marker based pagination to get items. 
 With marker based pagination you can iterate over folders containing more than 300000 elements.
-If you want to use PagingParameters to start marker based pagination: 
+With marker based pagination you cannot use sorting. Use `SortParameters.none()` turn off sort.
```
arjankowski, line 132: to turn off

arjankowski (dismissed):

arjankowski (approved):

PR #1078, opened 2022-08-01
feat: Added protected getter for `BoxAPIRequest#sslSocketFactory`

bartlomiejleszczynski:
As in https://jira.inside-box.net/browse/SDK-2417 - added a getter to sslSocketFactory to make easier usage during testing.


antusus:
We will go with a different solution.

PR #1079, opened 2022-08-04
fix: buffered body write and fixed SDK logging

antusus:


lukaszsocha2 (dismissed):

lukaszsocha2 (approved):
Looks ok. Please remember that removing public method `enableLogger()` is small breaking change and we need to communicate it proper way

PR #1083, opened 2022-08-16
feat: Using `OkHttp` in Java SDK

antusus:
This is a POC of replacing current Java 8 HTTP client with pure OkHttp client. I mean pure since I'm not dealing now with retries - so they are handled the same way.

In this POC I'm not adding any async code to Java SDK at this will be bery complicated. We would need to deal with the fact that when we want to get authorization token it will be a future that at some point in time will be resolved. The OkHttp client supports asynchronous HTTP operations and it is easy to return `CompletableFuture` from request.

# Challenges
## SslSocketFactory
In the old code we were setting SslSocketFactory on every request. Now this must be done on `BoxAPIConnection`. This is a breaking change since we know users are using reflection to override this setting.

## Mutable `BoxAPIConnection`
~~OhHttp client is immutable and current `BoxAPIConnection` is not. We can deal with this with replacing OkHttp client each time some method that changes its behaviour is called - like adding a proxy. We can do this like so:~~
```java
// copy existing client into the builder
OkHttp.Builder builder = new OkHttpClient.Builder(this.httpClient);
// now we can modify client behaviour
```
~~Or we can do a breaking change and make our `BoxAPIConnection` immutable as well.~~
For now we are not going to make `BoxAPIConnection` immutable.

## `RequestInterceptor`
Request interceptor allows users to modify `BoxAPIRequest` request. Since we are building OkHttp Request after applying  interceptor we should be fine as all changes are reflected on the final request.

## `AbstractBoxMultipartRequest` and retrying
~~In our SDK we are using `InputStream` to send files to box. This might be a cause of errors when retrying uploads. We are reading the stream containing the file content and converting it to OkHttp `RequestBody`. However once the stream is read and we would like to try again the input stream with file is exhausted. We will have to reset its state somehow to allow for retries.~~
Postpone - this works the same way as it was before.

## `BoxFileUploadSession`
The `BoxFileUploadSession` basically sends multiple multiform requests that are splitting file to parts. This should be doable. Same as with other uploads there is as outstanding issue with exhausted `InputStream`.

## Redirects
~~When uploading a file Box returns new URL where file must be uploaded. The `BoxAPIConnection` was handling that but the OkHttp client can do that as well. We have to figure out which way to go.
With current solution there are warnings reported by OkHttp:
`A connection to https://dl.boxcloud.com/ was leaked.` which suggest we have to solve redirect problems.~~
Solved - OK Http handles redirects. Another client that is not following redirects was added and used when we do not want to follow them but extract data from response.

## Proxy
~~OkHttp supports way more proxies than we do. We have to prepare new abstraction for setting proxy.~~
Done.

## Closing connection
~~`BoxAPIResponse` holds response from OK HTTP as it contains stream of data. When this data is consumed by reading body or writing it to destination (binary data) we are closing the stream and it should unlock HTTP connection. However what if user would not read response body and ignore it. Then warnings from OK HTTP will appear about leaked connections. Maybe we should add this info to documentation. Or maybe `BoxAPIResponse` should implement `Closable` interface and this might help users to close stream automatically.~~
`BoxAPIResponse` implements `Closable`. Added methods that do not expose BoxAPIResponse to the user or any stream. We want users to either receive JSONs or provide a stream to which binary content will be written to.

## HTTP2
By default for HTTPs requests there is a protocol negotiation. Client will check if the HTTP2 protocol can be used and if not fallback to HTTP1.1 (HTTP1.0 is not supported, nor it is supported in Box). 
Now we could introduce a way to force SDK to only use HTTP2 and skip negotiation step:
 - `BoxAPIConnection#forceHttp2Protocol()` - to make SDK use only HTTP2 protocol
 - `BoxAPIConnection#enableHttpProtocolNegotiation()`  - default setting that tries HTTP2 but can downgrade to HTTP1.1
 - `BoxAPIConnection#protocols(HttpProtocol protocol...)` - and create an evolving Enum that will match what OkHttp library enables clients to set. This way you can set just `HTTP2_PRIOR_KNOWLEDGE` (to force HTTP2) or `HTTP2` and `HTTP1.1` to allow for protocol negotiation.

## Things to remove
 - `BatchAPIRequest` this deprecated class will not be converted


src/main/java/com/box/sdk/BoxUser.java
```diff
@@ -680,12 +682,10 @@ public BoxFolder.Info transferContent(String destinationUserID) {
      *
      * @return InputStream representing the user avater.
      */
-    public InputStream getAvatar() {
+    public void downloadAvatar(OutputStream outputStream) {
```
antusus, line 685: Since response holds opened connection if user forgets to close it then we are leaking http connections. It is better to use same approach we have with files download where user provides an output stream, we write there and then we can close connection and prevent leaking.

lukaszsocha2 (approved):
Good job 🚀

PR #1086, opened 2022-08-23
feat: Adding Jackson to Java SDK

antusus:
All important data is stored in `Info` classes for example `BoxFile.Info`. Info classes are not static and cannot exist without resource classes. Resource classes contains API connection and String ID. There is no possibility to deserialise them. 

It makes little sense to introduce Jackson to SDK as we will gain nothing from its features.


antusus:
Created for testing purpouses

PR #1087, opened 2022-08-23
feat: Add `content-type` sign request response

congminh1254:
Closes: SDK-2253

### DIFF
--- a/src/main/java/com/box/sdk/BoxSignRequestSigner.java
+++ b/src/main/java/com/box/sdk/BoxSignRequestSigner.java
@@ -338,6 +338,94 @@ public class BoxSignRequestSigner extends BoxJSONObject {
     }
 
     /**
+     * Represents a content type of input.
+     */
+    public enum BoxSignRequestInputContentType {
+        /**
+         * Initial content type
+         */
+        Initial("initial"),
+        /**
+         * Stamp content type
+         */
+        Stamp("stamp"),
+        /**
+         * Signature content type
+         */
+        Signature("signature"),
+        /**
+         * Company content type
+         */
+        Company("company"),
+        /**
+         * Title content type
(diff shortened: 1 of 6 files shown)


src/main/java/com/box/sdk/BoxSignRequestSigner.java
```diff
@@ -337,6 +337,94 @@ static BoxSignRequestInputType fromJSONString(String jsonValue) {
...
+        /**
+         * Checkbox content type
+         */
+        Checkbox("checkbox");
+
+        private final String jsonValue;
+        BoxSignRequestInputContentType(String jsonValue) {
+            this.jsonValue = jsonValue;
+        }
+        static BoxSignRequestInputContentType fromJSONString(String jsonValue) {
```
antusus, line 397: Maybe let's use switch which is not used in SDK and I have no idea why...
And we should provide details about the faulty value

```java
static BoxSignRequestInputContentType fromJSONString(String jsonValue) {
    switch (jsonValue) {
        case "initial":
            return Initial;
        case "stamp":
            return Stamp;
        case "signature":
            return Signature;
        case "company":
            return Company;
        case "title":
            return Title;
        case "email":
            return Email;
        case "full_name":
            return FullName;
        case "first_name":
            return FirstName;
        case "last_name":
            return LastName;
        case "text":
            return Text;
        case "date":
            return Date;
        case "checkbox":
            return Checkbox;
        default:
            throw new IllegalArgumentException(
                format("The provided JSON value '%s' isn't a valid BoxSignRequestInputContentType.", jsonValue)
            );
    }
}
```

antusus (commented):
Let's improve this a little

antusus (approved):
👍

PR #1095, opened 2022-09-05
fix: Stop sending not specified optional fields when creating a user

lukaszsocha2:
Closes: SDK-2589

### DIFF
--- a/src/main/java/com/box/sdk/CreateUserParams.java
+++ b/src/main/java/com/box/sdk/CreateUserParams.java
@@ -1,20 +1,21 @@
 package com.box.sdk;
 
+import java.util.HashMap;
 import java.util.Map;
 
 /**
  * Contains optional parameters for creating a new enterprise user on Box.
  */
 public class CreateUserParams {
-    private boolean canSeeManagedUsers;
-    private boolean isExemptFromDeviceLimits;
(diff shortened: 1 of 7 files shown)


src/main/java/com/box/sdk/CreateUserParams.java
```diff
@@ -28,7 +29,7 @@ public class CreateUserParams {
      *
      * @return true if the new user will be able to see other enterprise users in their contact list; otherwise false.
      */
-    public boolean getCanSeeManagedUsers() {
+    public Boolean getCanSeeManagedUsers() {
```
antusus, line 32: This is a breaking change in the API we have to mark it in the PR title

antusus (changes requested):
We are making a breaking change in the API because of the CreateUserParams are now returning null. We have to show it in the title

src/main/java/com/box/sdk/CreateUserParams.java
lukaszsocha2: I consider it as a small breaking change (Usually clients won't use getters for an object used only to make a request). Would adding `!` not cause bumping major version ?

antusus (approved):
👍

antusus (dismissed):
👍

antusus (approved):
👍

PR #1098, opened 2022-09-08
fix: Add missing constructor to `BoxNotificationEmail` class

lukaszsocha2:
Closes: SDK-2607


src/main/java/com/box/sdk/BoxNotificationEmail.java
```diff
@@ -39,6 +41,8 @@ public BoxNotificationEmail(String json) {
      */
     BoxNotificationEmail(JsonObject jsonObject) {
         super(jsonObject);
+        this.email = jsonObject.getString("email", null);
```
lukaszsocha2, line 44: why these ? It will be set anyway

antusus (approved):
👍

PR #1109, opened 2022-10-07
feat: Added Box Sign webhooks

antusus:
For now Webhooks can be created on file or folder. Later there will support Sign Requests


mhagmajer (approved):
👌

PR #1118, opened 2022-11-08
feat: Adding proxy support

antusus:


src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -231,20 +234,47 @@ private void buildHttpClients() {
...
+            if (proxyUsername != null && proxyPassword != null) {
+                builder.proxyAuthenticator((route, response) -> {
+                    String credential = Credentials.basic(proxyUsername, proxyPassword);
+                    return response.request().newBuilder()
+                        .header("Proxy-Authorization", credential)
+                        .build();
+                });
+            }
+            if (this.authenticator != null) {
+                builder.proxyAuthenticator(authenticator);
```
antusus, line 254: Question: Right now if user wants to use basic proxy authenticator and then provides its own it it replaced with his own. I'm not throwing any exception just overriding it. Maybe this should be an exception? Maybe User cannot call `setProxyAuthenticator` and `setProxyBasicAuthentication` (or `setProxyPassword` along with `setProxyUsername`)?

lukaszsocha2 (approved):
lgtm

Issue #1160, opened 2023-04-17
Runtime Exception while searching for a User with SDk


pbale:
- [ yes] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [ yes] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [yes ] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [ yes] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
Runtime exception is thrown while searching for a user. I tried the search on Postman with URL, https://api.box.com/2.0/users?filter_term=gad011&fields=id,status,login and it did returned me the successful payload. but when I use the box java SDK its failing. Any help will be greatly appreciated. If you want to see more information on this please refer to the case number #2832801 on Box SDK forums. 
Note: My application uses Spring boot Framework.

### Steps to Reproduce
1. Get oAuthCode and save token information into the database
2. Check if the token is expired by, (currentTime-lastRefresh > expired) perform BoxAPIConnection.refresh()
3. Search for the user(gad011) 
Iterables.get(
            Iterables.filter(BoxUser.getAllEnterpriseUsers(BoxAPIConnection, userId, fields), onLoginFilter), 0);
where userId = gad011, fields = id, status, created_at,modifies_at, login
4. step 3 produces RuntimeException:Null

### Expected Behavior
I am expecting to see the user information with the specified fields.

### Error Message, Including Stack Trace

63248 [http-nio-8080-exec-1] ERROR o.a.c.c.C.[.[.[.[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.RuntimeException] with root cause 
   java.lang.RuntimeException: null
	at com.box.sdk.BoxAPIRequest.send(BoxAPIRequest.java:472)
	at com.box.sdk.BoxAPIRequest.send(BoxAPIRequest.java:404)
	at com.box.sdk.JsonIterator.loadNextPage(JsonIterator.java:68)
	at com.box.sdk.JsonIterator.loadNextJsonObject(JsonIterator.java:102)
	at com.box.sdk.JsonIterator.hasNext(JsonIterator.java:29)
	at com.box.sdk.BoxUserIterator.hasNext(BoxUserIterator.java:21)
[...]

PR #1161, opened 2023-04-20
fix: Fallback to default value of `maxRetryAttempts` when restoring `BoxAPIConnection`

lukaszsocha2:
Closes: SDK-3109
Fixes: #1160


antusus (approved):
👍

Issue #1183, opened 2023-06-26
Uploading a new version to a target box file using box source file not working after SDK upgrade


hmousaiti:
- [x] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [x] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [x] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
1.	App user is an owner of a folder, and also is a collaborator with role “Viewer” in another user’s file.
2.	App user is trying to download the file from the user’s box account and upload it as new version to the owned folder.
3.	The download and upload process happen in parallel and in an asynchronous way, downloading the file from box and writing it to an output stream, meanwhile uploading the downloaded part using an input stream connected to the output stream.

The above behavior is being done using PipedOutputStream, and PipedInputStream. It is working fine using Box SDK 3.3.0. But, when we migrated to use SDK 4.0.1, we started to get and exception that:

java.util.concurrent.CompletionException: java.lang.RuntimeException: java.io.IOException: Pipe closed
	at org.springframework.aop.interceptor.AsyncExecutionAspectSupport.lambda$doSubmit$3(AsyncExecutionAspectSupport.java:281) ~[spring-aop-5.3.20.jar:5.3.20]
	at org.springframework.aop.interceptor.AsyncExecutionAspectSupport$$Lambda$1300.000000001A83A8F0.get(Unknown Source) ~[na:na]

### Steps to Reproduce

1. Spring boot application which is downloading a file from box and upload that downloaded file as a new version to another target file.
2. Given PipedOutputStream out, and PipedInputStream in = new PipedInputStream(8192).
3. in.connect(out);
4. Download the file to the "out"
5. Upload the new version from the "in"

### Error Message, Including Stack Trace
java.util.concurrent.CompletionException: java.lang.RuntimeException: java.io.IOException: Pipe closed
	at org.springframework.aop.interceptor.AsyncExecutionAspectSupport.lambda$doSubmit$3(AsyncExecutionAspectSupport.java:281) ~[spring-aop-5.3.20.jar:5.3.20]
	at org.springframework.aop.interceptor.AsyncExecutionAspectSupport$$Lambda$1300.000000001A83A8F0.get(Unknown Source) ~[na:na]
[...]

PR #1189, opened 2023-08-02
fix: Fixed upload when data are coming from a dynamic source

antusus:
With OkHttp not really working with stream we have our own solution. But it can work with them as long as we tell it that content size if unknown.

This was the source of the problem in the #1183 where two threads started and one was feeding data to other. First was downloading a file second was uploading new version.
This should also address failing uploads when dealing with a stream which full size cannot be determined upfront (#1190 ).

### DIFF
--- a/src/main/java/com/box/sdk/RequestBodyFromStream.java
+++ b/src/main/java/com/box/sdk/RequestBodyFromStream.java
@@ -41,7 +41,7 @@ final class RequestBodyFromStream extends RequestBody {
     }
 
     /**
-     * Returns content length. If the content length cannot be derermined
+     * Returns content length. If the content length cannot be determined
      * (is coming from a stream) this value will be -1.
      * @return Long representing content length
      */
--- a/src/main/java/com/box/sdk/RequestBodyFromStream.java
+++ b/src/main/java/com/box/sdk/RequestBodyFromStream.java
@@ -5,6 +5,8 @@ import java.io.InputStream;
 import okhttp3.MediaType;
 import okhttp3.RequestBody;
 import okio.BufferedSink;
+import okio.Okio;
+import okio.Source;
 
 /**
  * Utility class to write body from stream to BufferedSink used by OkHttp.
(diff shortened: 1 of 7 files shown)


src/main/java/com/box/sdk/RequestBodyFromStream.java
```diff
@@ -13,41 +15,70 @@ final class RequestBodyFromStream extends RequestBody {
...
+    public void writeTo(BufferedSink bufferedSink) {
+        if (progressListener == null) {
+            writeWithoutProgressListener(bufferedSink);
+        } else {
+            writeWithProgressListener(bufferedSink);
+        }
     }
 
+    /**
+     * Returns content length. If the content length cannot be derermined
```
lukaszsocha2, line 44: determined

src/main/java/com/box/sdk/RequestBodyFromStream.java
```diff
@@ -13,41 +15,70 @@ final class RequestBodyFromStream extends RequestBody {
...
+    private void writeWithProgressListener(BufferedSink bufferedSink) {
+        try {
+            byte[] buffer = new byte[AbstractBoxMultipartRequest.BUFFER_SIZE];
+            int n = this.inputStream.read(buffer);
+            int totalWritten = 0;
+            while (n != -1) {
+                bufferedSink.write(buffer, 0, n);
+                totalWritten += n;
+                if (progressListener != null) {
+                    progressListener.onProgressChanged(totalWritten, this.contentLength());
```
lukaszsocha2, line 70: What if `contentLength()` return -1. Would progress listener work correctly ?

antusus: I've added details here: https://github.com/box/box-java-sdk/blob/24b17565570d032f93752036a9d1ae6cfb8ce8f4/src/main/java/com/box/sdk/ProgressListener.java#L10

in short total size will be -1 but bytes written should be properly returned.

lukaszsocha2 (approved):
Maybe close the issue with this pr

PR #1241, opened 2024-03-28
fix: Support create empty shared link

congminh1254:


antusus (dismissed):
👍

src/test/java/com/box/sdk/BoxSharedLinkTest.java
```diff
@@ -43,7 +42,7 @@ public void doNotStorePendingChangeWhenPermissionsNotChanged() {
         link.setPermissions(this.permissions(true, false));
 
         assertThat(link.getPermissions(), is(permissions(true, false)));
-        assertThat(link.getPendingChangesAsJsonObject(), is(nullValue()));
+        assertThat(link.getPendingChangesAsJsonObject().toString(), is("{}"));
```
antusus, line 45: But now even when there are no changes we will do an update on the object. Is that OK?

congminh1254: Regarding the [docs](https://developer.box.com/reference/put-files-id--remove-shared-link/), if the `shared_link` is null, so the share link will be removed, so I think it will better anyway to return empty object instead of null.

antusus (approved):
👍

PR #1287, opened 2025-02-11
feat: Support `zstd` encoding for downloads

congminh1254:
Support use zstd encoding while downloading files.

### DIFF
--- a/src/intTest/java/com/box/sdk/BoxFileIT.java
+++ b/src/intTest/java/com/box/sdk/BoxFileIT.java
@@ -1,7 +1,9 @@
 package com.box.sdk;
 
+import static com.box.sdk.BinaryBodyUtils.writeStream;
 import static com.box.sdk.BoxApiProvider.jwtApiForServiceAccount;
 import static com.box.sdk.BoxFile.ALL_VERSION_FIELDS;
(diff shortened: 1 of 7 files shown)


src/main/java/com/box/sdk/BoxAPIConnection.java
```diff
@@ -122,6 +122,7 @@ public class BoxAPIConnection {
     private int maxRetryAttempts;
     private int connectTimeout;
     private int readTimeout;
+    private Boolean useZstdCompression;
```
lukaszsocha2, line 125: Why did you use Boolean, not boolean ?

src/intTest/java/com/box/sdk/BoxFileIT.java
```diff
@@ -225,6 +227,64 @@ public void uploadAndDownloadFileSucceeds() throws IOException {
 
     }
 
+    @Test
+    public void uploadAndDownloadFileUseZstdSucceeds() throws IOException {
```
lukaszsocha2, line 231: maybe change the name to downloadFileUseZstdSucceeds, actually upload is just precondition to download file and this name suggests that we upload file using zstd as well, which is not the case

lukaszsocha2 (approved):
LGTM

congminh1254:
It's gtg 🍡

PR #1292, opened 2025-02-24
fix: Fix parsing content length header

congminh1254:


src/test/java/com/box/sdk/BoxAPIResponseTest.java
```diff
@@ -85,4 +144,18 @@ public void testJsonResponseToString() {
                 + "Body:\n{\"foo\":\"bar\"}")
         );
     }
+
+    @Test
+    public void testLargeBinaryResponseContentLength() {
+        long contentLength = 2352815193L;
```
antusus, line 150: What is this magic number? You want it to be larger than MAX int? Some random value?

congminh1254: This is a number bigger than Max Int and which is customer having issue with, can be also considered as a random number :/

antusus (approved):
👍

Issue #1305, opened 2025-04-26
BoxDeserializationException for Turkish language


Nahian98:
### Description of the Issue
I am getting the following error message after changing my app locale to Turkish. 

```
Deserialization failed on: [ "field name": status | "field value": "active" ] 
at com.box.sdk.BoxUser$Info.parseJSONMember(BoxUser.java:1443) 
at com.box.sdk.BoxJSONObject.update(BoxJSONObject.java:205) 
at com.box.sdk.BoxJSONObject.<init>(BoxJSONObject.java:56) 
at com.box.sdk.BoxResource$Info.<init>(BoxResource.java:189) 
at com.box.sdk.BoxCollaborator$Info.<init>(BoxCollaborator.java:113) 
at com.box.sdk.BoxUser$Info.<init>(BoxUser.java:915) 
at com.box.sdk.BoxUser.getInfo(BoxUser.java:470) 
at com.offlinemusicplayer.view.activity.HomeActivity$setupBoxAPI$2.invokeSuspend(HomeActivity.kt:916) 
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) 
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100) 
at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:113) 
at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89) 
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586) 
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:820) 
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717) 
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704) Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@5a69899, 

Dispatchers.Main.immediate] Caused by: java.lang.IllegalArgumentException: No enum constant com.box.sdk.BoxUser.Status.ACTİVE 
at java.lang.Enum.valueOf(Enum.java:300) 
at com.box.sdk.BoxUser$Status.valueOf(BoxUser.java:832) 
[...]

PR #1306, opened 2025-04-28
fix: use `Locale.ROOT` to prevent issues with non-US locales

mwwoda:
Fixes https://github.com/box/box-java-sdk/issues/1305

- Use Locale.ROOT in `toLowerCase` and `toUpperCase` to prevent unwanted conversion on characters on turkish locale (see comment for https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/String.java#L3743)
- Fix github actions 322 code error by bumping version of used actions. Issue was most likely caused by using outdated version of https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/ somewhere in our actions

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -15,9 +15,9 @@ jobs:
     name: Java ${{ matrix.java }} (${{ matrix.distribution }})
     steps:
       - name: Checkout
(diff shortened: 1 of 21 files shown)


antusus (approved):
👍

PR #1315, opened 2025-06-24
fix: Fix compare message webhook message signature

congminh1254:


Pask423 (approved):
Looks good, one liner not much to review

PR #1910, opened 2026-07-07
deps: Upgrade `Jackson` libraries to 2.21.4

arjankowski:
Updated Jackson dependencies to remediate reported jackson-databind CVEs. This uses the fixed Jackson BOM-compatible set: jackson-annotations:2.21, jackson-core:2.21.4, and jackson-databind:2.21.4

Also updated the coverage upload job because the existing coverallsJacoco Gradle task was failing with HTTP 400 after the JaCoCo report was generated successfully. The workflow now uploads the JaCoCo XML report via the official Coveralls GitHub Action from a single matrix job

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -29,8 +29,12 @@ jobs:
       - name: Build and test
         run: ./gradlew check --stacktrace
       - name: Generate JaCoCo report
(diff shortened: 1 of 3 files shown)


antusus (approved):
👍

lukaszsocha2 (approved):
lgtm

PR #1958, opened 2026-08-20
fix: avoid null chunk when file size is a multiple of part size

### LINKED ISSUE #1957: BUG in Chunked Upload for Big Files
### Description of the Issue

When uploading a file larger than 50MB, Chunked Uploads should be used, however, if the file size is exactly multiples of 8MB then a NullPointerException is thrown.

### Steps to Reproduce

1. Create a file that is exactly 64MiB sized locally
2. Upload to box using BoxClient.chunkedUpload.uploadBigFile(inputStream, fileName, fileSize, folderId) 
3. NullPointerException will be thrown.

### Expected Behavior

Expect to upload file successfully.

### Issue Found in Version

box-java-sdk 10.16.0
In com.box.sdkgen.internal.utils.UtilsManager, in the method iterateChunks
if a chunk exactly divisible by 8MB then it returns a null InputStream

### Proposed fix

public static Iterator<InputStream> iterateChunks(
      InputStream stream, long chunkSize, long fileSize) {
	  return new Iterator<InputStream>() {

	        private InputStream nextChunk;
	        private boolean initialized = false;

	        private void prepareNext() {
	            if (initialized) {
	                return;
	            }

	            initialized = true;

	            try {
	                byte[] buffer = new byte[(int) chunkSize];
	                int bytesRead = 0;

	                while (bytesRead < chunkSize) {
	                    int read = stream.read(
	                        buffer,
	                        bytesRead,
	                        (int) (chunkSize - bytesRead)
	                    );

	                    if (read == -1) {
	                        break;
	                    }

	                    bytesRead += read;
	                }

	                if (bytesRead == 0) {
	                    nextChunk = null;
	                    return;
	                }

	                nextChunk = new ByteArrayInputStream(buffer, 0, bytesRead);

	            } catch (IOException e) {
	                throw new RuntimeException("Error reading from stream", e);
	            }
	        }

	        @Override
	        public boolean hasNext() {
	            prepareNext();
	            return nextChunk != null;
	        }

	        @Override
	        public InputStream next() {
	            prepareNext();

	            if (nextChunk == null) {
	                throw new NoSuchElementException();
	            }

	            InputStream result = nextChunk;

	            nextChunk = null;
	            initialized = false;

	            return result;
	        }
	    };
  }

antusus:
`iterateChunks` treated `hasNext` as true until EOF, so an exact-size last part was followed by a null stream and an NPE in `uploadBigFile`.

Closes #1957

### DIFF
--- a/src/main/java/com/box/sdkgen/internal/utils/UtilsManager.java
+++ b/src/main/java/com/box/sdkgen/internal/utils/UtilsManager.java
@@ -34,6 +34,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


antusus:
WIll need to update the codegen first

PR #1966, opened 2026-08-26
fix(boxsdkgen): avoid null chunk when file size is a multiple of part size (box/box-codegen#980)

box-sdk-build:


KwiatkowskiML (approved):
LGTM

